PR #23, opened 2026-07-11
fix(compress): per-content-type keep policy so the digest keeps a test run's verdict

### LINKED ISSUE #22: Tier-1 digest drops a passing test run's verdict (content-blind keep rule)
### Summary
distil drops key lines at times which affects assessment. Typically this might be test results, where distil removes the actual results line causing more loops. See more details [here](https://moonlit-piroshki-fbaa04.netlify.app/)

### Problem
The Tier-1 reversible digest (`distil/compress/tier1.py`) keeps head/tail context plus a generic net (`DECISION:` markers and `error|exception|traceback|fail|warn|...` lines), but nothing recognizes a **result-summary** line. On an all-passing test log the summary matches no keep rule, and `tail=1` misses the multi-line footer, so the digest folds the one decision-relevant line behind a handle.

### Repro (facts)

A full `vitest run` (1955 tests, ~22,971 tokens at o200k) digests ~90%, but:
- `Tests  1955 passed (1955)` → **dropped** (folded behind a handle)
- incidental `ERROR [Crypto] ...` stdout from the code under test → **kept**

For a "validate tests" task the pass/fail summary is the whole payload, so a large (technically reversible) reduction that hides it is a quality miss.

### Note

The `tier1.py` docstring already anticipates the fix: *"in production this is where a learned per-content-type codec or a salience model plugs in."*

### Fix direction (PR upcoming)

A model-free per-content-type keep policy: LOG pins the result summary, TRACEBACK pins stack frames, DIFF pins hunk headers; each inherits the generic net. Stays non-inferior on the corpus (`make gate`) and byte-reversible. PR to follow.

—
Filed by Claude Opus 4.8 assisting @PliablePixels.

pliablepixels:
Fixes #22.

## Problem

The Tier-1 digest (`compress/tier1.py`) keeps head/tail plus a generic net (`DECISION:` and `error|fail|warn|...`), but no rule recognizes a **result-summary** line, and `tail=1` misses a multi-line footer. So an all-passing test log's verdict gets folded behind a handle.

## Fix

- **`compress/keep_policy.py`** (new): `classify(text) → {generic, log, traceback, diff}` and `must_keep(line, kind)`. LOG pins the result summary, TRACEBACK pins stack frames, DIFF pins hunk/file headers; every kind inherits the generic net. Model-free and deterministic — the "per-content-type codec" the `tier1.py` docstring names.
- **`digest()`** classifies once and applies the policy. Markers become `<< +N lines omitted, handle=H >>`, with a `(E error, W warn, O other)` breakdown appended only when flagged lines are folded (usually none, since the keep policy already surfaces them). Recovery handle and byte-exact restore unchanged.

## Before / after (the repro from #22)

Same `vitest run` log (1955 tests, ~22,971 tokens at o200k):

| | verdict `1955 passed` | marker |
|---|---|---|
| `main` | **dropped** | `<< +1 lines, handle=965d42f6 >>` |
| this PR | **retained** | `<< +1 lines omitted, handle=965d42f6 >>` |

The digest still reduces ~89% (22,971 → 2,515 o200k), now keeping the decision-relevant line.

## Corpus impact (`make bench`, $ saved per domain)

| domain | `main` | this PR | Δ |
|---|---|---|---|
| ops/sre | 32.8% | 32.6% | −0.2 |
| coding | 27.0% | 25.1% | −1.9 |
| support | 32.6% | 32.3% | −0.3 |
| research | 25.7% | 25.5% | −0.2 |
| data-analysis | 18.1% | 18.0% | −0.1 |
| devops | 22.8% | 22.6% | −0.2 |
| finance | 24.9% | 24.8% | −0.1 |
| **aggregate** | **25.5%** | **25.1%** | **−0.4** |

A small, bounded cost for keeping the load-bearing lines; coding-bugfix moves most since it is richest in log/trace/diff content.

## Gate & checks

- `make gate`: **PASS** — every trajectory certified non-inferior, aggressive rejected; byte-fidelity **PASS** (reversible + append-only across all 7 domains).
- `make lint` (ruff) ✓ · `mypy` ✓ (86 files) · full suite ✓ (+16 tests in `tests/test_keep_policy.py`).
- Coverage: `keep_policy.py` 100%, `tier1.py` 100%; the change is coverage-positive. (The repo's overall floor sits ~94.8% locally on `main` too — pre-existing, not introduced here.)

## Docs

ADR `docs/adr/0002-content-type-keep-policy.md`; `tier1.py` docstring and marker-format descriptions updated. No CHANGELOG entry (curated at release per `RELEASING.md`).

—
Authored by Claude Opus 4.8 assisting @PliablePixels.

### DIFF
--- a/distil/compress/keep_policy.py
+++ b/distil/compress/keep_policy.py
@@ -1,0 +1,114 @@
+"""Per-content-type keep policy for the Tier-1 reversible digest.
+
+The digest folds a verbose block's droppable middle behind a handle. WHICH lines
+are droppable depends on the content: a passing test run's value is its pass/fail
+summary, a traceback's is its frames, a diff's is its hunk headers. This module
+classifies a block and returns the per-type "must keep" rule, so the digest never
+drops the line the content is actually about. It also provides the by-category
+count that goes in the digest marker, so the agent can judge whether a folded span
+is worth expanding.
+
+Model-free and deterministic. The keep rule is biased toward over-keeping: a stray
+kept line costs a little compression, a missed one costs the agent the decisive
+line. This is the "per-content-type codec" the tier1 docstring anticipated.
+"""
+
+from __future__ import annotations
+
+import re
+from enum import Enum
+
+
+class ContentKind(str, Enum):
+    GENERIC = "generic"
+    LOG = "log"  # test / build / CI console output
+    TRACEBACK = "traceback"
+    DIFF = "diff"
+
+
+# GENERIC salience net (unchanged from the original digest): explicit DECISION
+# markers plus the lines an agent most often needs verbatim to react.
+_GENERIC_RE = re.compile(r"error|exception|traceback|fail|warn|panic|fatal", re.IGNORECASE)
+
+# LOG result-summary lines a runner prints. Anchored on a count next to an outcome
+# word, so per-test lines ("(3 tests) 5ms") are NOT kept, only summaries
+# ("1955 passed", "2 failed", "3 skipped"), plus common footer labels and exit codes.
+_RESULT_RE = re.compile(
+    r"\d+\s+(passed|passing|failed|failing|skipped|pending|errored)\b"
+    r"|^\s*(Test Files|Tests|Duration|Start at)\b"
+    r"|test result:"  # cargo
+    r"|=+.*\b(passed|failed|error)\b.*=+"  # pytest summary rule
+    r"|\bexit(?:\s|-)?code\b",
+    re.IGNORECASE,
+)
+
+# TRACEBACK: keep every stack frame so the call site survives, not just the message.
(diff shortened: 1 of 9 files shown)


dshakes:
adopted this in 1.15.0 — thanks, it's the right shape and it's the keep decision now.

quick note on how it landed: the branch conflicted because a first-pass verdict fix already went out in 1.14.1, and 1.15.0 added a dedup layer + a bench check on top, so a straight merge would've reverted those. so instead of merging as-is I pulled your `keep_policy.py` in — classify into log/traceback/diff/generic + the per-kind keeps, your result regex became the superset — and folded the dedup + check back over it. it's in 7a8d7ed with you as co-author so the credit's in the history, and the site docs describe it now. it also ended up being the base for a query-aware keep layer on top.

closing as adopted rather than merged, just to dodge that destructive-merge problem — but the work's in main. appreciate the clean module and the writeup.

Issue #26, opened 2026-07-12
Feature request: distil dissect — per-session deep-dive report (working implementation available)


pliablepixels:
[foo.html](https://github.com/user-attachments/files/29944829/foo.html)

### The gap

distil records rich local telemetry (the savings ledger, restore store, shadow verdicts), but every reader is an **aggregate**: `leaderboard`/`dashboard` roll up all sessions, `shadow-stats` is a global window. There is no way to answer the question users actually ask after a session:

> *"What exactly happened to **that** session — what was compressed, what wasn't, what did it cost me, and did anything go wrong?"*

Today that takes hand-written `jq` over `savings.jsonl`, grepping `restore/`, and guesswork. In practice it also means real failures stay invisible — we only found #25 (expand escapes + silent shadow) by manually spelunking those files.

### Proposal

A new subcommand, following the existing reader conventions (`--json`, `--html`, plain text, stdlib-only):

```
distil dissect                    # list sessions (tool, times, saved%) — interactive picker on a TTY
distil dissect latest             # deep-dive one session (id, unique prefix, or `latest`)
distil dissect latest --html p.html
distil dissect --serve            # localhost portal: / = picker, /session/<sid> = live report
```

The report leads with a plain-language **"What happened"** narrative (for readers who won't hover a tile), then drills down:

- **Savings**: baseline → actually-sent input tokens, per model; dollars flagged *notional* on flat-rate billing, with a "context budget stretched ~N×" headroom figure instead.
- **Savings by mechanism**: digest folds vs cache-delta.
- **What was NOT optimized and why**: system-prompt + tool-definition overhead (resent verbatim every request), verbatim requests, unbooked (failed/retried) requests.
- **Digest inventory**: folded blocks by content kind (`log:l`, `prose:m`, …), largest folds, re-fold churn, and whether each block is still recoverable from `restore/`.
- **Ground truth**: billed `usage.input/output_tokens` captured from API responses (JSON and SSE), with a calibration figure showing how close the heuristic estimate came.
- **Latency by path** — including the `--expand` forced-buffering TTFT tax, measured.
- **Quality loops**: expand resolutions, shadow samples/verdicts.
- **A "worth your attention" anomaly list**: proxy bypass, shadow configured-but-silent, expand-never-intercepted (both #25 signatures detected automatically), unbooked-request spikes, savings far below the user's own median.

[...]

Issue #25, opened 2026-07-11
wrap --expand: distil_expand tool_use escapes to the client on streamed no-new-handle requests; --shadow silently records nothing (subscription/OAuth session)


pliablepixels:
In a real `distil wrap --expand --session-delta --shadow 0.1 -- claude` session (Claude Code, subscription/OAuth), compression worked end-to-end (19 requests, 1,819,796 → 627,955 input tokens, 65.5%, subagent traffic included), but **both quality-assurance features in that flag set showed no signs of life**:

1. `--expand`: the agent called the advertised `distil_expand` tool and the tool_use **escaped to the client unresolved** — Claude Code answered `Error: No such tool available: distil_expand`. `~/.distil/expand-stats.json` after the session: 1,535 blocks digested, **0 expanded**. So the lossy digest currently has no working recovery path in this configuration.
2. `--shadow 0.1`: after the same 19 requests, `distil shadow-stats` reports "No shadow samples yet" and no shadow ledger file exists in `~/.distil` at all.

## Environment

- distil 1.14.1 (installed binary matches the v1.14.1 tag)
- macOS (Darwin 25.5.0), zsh
- Agent: Claude Code (Anthropic Messages API, **streaming**), subscription billing (OAuth, no API key)
- Launch: `alias claude='distil wrap --expand --session-delta --shadow 0.1 -- claude'`

## Symptom 1 — `distil_expand` tool_use relayed raw to the client

What the agent saw, mid-session (conversation full of live digest markers from earlier turns):

- Tool results contained markers like `<< +25 lines, handle=6014b735 >>`, and the `distil_expand` tool definition was visible to the model — so `inject_expand_tool` fired on earlier requests.
- The model then called `distil_expand` with `{"handle": "6014b735"}` (a valid handle from an earlier turn) and the client returned `No such tool available: distil_expand` — i.e. the tool_use block reached Claude Code instead of being resolved by the proxy's transparent expand loop.

### Where I think it breaks (v1.14.1 sources)

Both the streaming bypass and the expand loop are gated on the **current request's** store having handles:

- `distil/proxy.py:604` — `if want_stream and not (expand and getattr(store, "handles", None)):` → streamed requests take the pass-through relay (`stream_upstream`, with `want_body=False`) whenever `store.handles` is falsy. Nothing inspects the streamed response for `distil_expand` tool_use blocks on this path.
[...]

PR #27, opened 2026-07-12
feat: distil dissect — per-session deep-dive report (implements #26)

pliablepixels:
Implements #26 — `distil dissect`: a per-session deep-dive report of what distil actually did, for any wrapped tool.

## What this adds

```
distil dissect                     # session picker (interactive on a TTY)
distil dissect latest              # one session: narrative + drill-down (--json / --html)
distil dissect --serve             # localhost portal: / picker, /session/<sid> live reports
distil dissect latest --transcript # opt-in: correlate with the agent's own conversation log
```

**The report** leads with a plain-language "What happened" narrative, then drills down: savings per model (dollars marked *notional* on flat-rate billing, with a headroom multiplier instead), savings by mechanism (digest vs cache-delta), what was *not* optimized and why (system-prompt/tool-definition overhead, verbatim and unbooked requests), the digest inventory (blocks by kind, largest folds, re-fold churn, restore-blob recoverability), billed usage captured from API responses with a heuristic-calibration figure, latency by path (the `--expand` buffering tax is now a measured number), quality loops, and a "worth your attention" anomaly list — which detects the #25 signatures automatically, so that class of silent failure can't hide again.

The HTML report is self-contained (inline SVG, no CDN, no new dependencies) with cursor-following tooltips that explain every distil term in plain English. A demo rendered from synthetic data is attached to #26.

**Transcript correlation** (strictly opt-in — it names tools/files/prompts, everything else stays content-free): joins the session with the agent's own log via the restore store's original bytes matched into transcript tool results. Yields folds named by producing tool + prompt, tools defined-but-never-invoked (on a real session: 20/24 tools ≈ 40k tokens/request paid for nothing), re-fetched-after-fold warnings (a multi-turn quality signal shadow sampling can't see), and cost per human turn.

Agent support is a pluggable adapter layer: `distil/transcripts/base.py` documents the contract (two methods — `discover`, `load` — normalizing into turns/tool-calls/tool-results); the registry keys on the wrap manifest's tool name. Claude Code is the first adapter; codex/gemini/aider are one module + one registry line each. Extensibility is proven by an in-test toy adapter.

## Proxy-side changes (small, fail-open, content-free)

Two logs, both under the existing `sessions/` TTL sweep, both wrapped so bookkeeping can never break a request:

- `sessions/<sid>.json` — wrap manifest (tool, argv, cwd, flags, billing, version), written once at startup.
- `sessions/<sid>.requests.jsonl` — one record per proxied request: token accounting from the extras the proxy already computes, per-block digest signatures (handle + kind:size + tokens, never content), billed `usage.*` (JSON parsed; SSE scanned from a bounded head/tail window in `streamrelay` via a new `usage_sink` param — nothing new is buffered), duration + stream-vs-buffered path, shadow/expand flags, per-tool definition sizes.

Pre-existing sessions degrade to a ledger-only view with an explicit note. The expand loop's `on_signal` seam additionally reports which handles were resolved (content-free).

## Conventions honored

- stdlib only, no new dependencies; full type hints; ruff/mypy clean.
- `make gate` + `distil verify` PASS on every commit (the branch is rebased onto v1.15.0, composing cleanly with the #25 fix — the report's forced-buffering metric now measures the TTFT trade your fix docstring mentions).
- No CHANGELOG entry, per RELEASING.md.
- 64 new tests in `tests/test_dissect.py` (proxy round-trips against a stub upstream incl. SSE usage capture, report math, renderers, portal routes, adapter parsing/discovery, correlation math, toy-adapter extensibility). Full suite: 1356 passed.

## Trying it

```
pipx install --force <this branch>   # or: uv run distil ...
distil wrap --expand --session-delta -- claude   # any base-url-honoring agent
distil dissect --transcript --serve              # then open http://127.0.0.1:8790/
```

## Review notes

The 10 commits are individually coherent if you prefer to review in stages (core+logging → insights → charts/UX → portal → transcript adapters). Equally happy to split this into that PR sequence instead — whatever reviews best. Naming and flag conventions are all up for adjustment.

Filed by Claude (Fable 5) assisting @pliablepixels

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/distil/cli.py
+++ b/distil/cli.py
@@ -302,7 +302,9 @@ def cmd_dissect(args: argparse.Namespace) -> int:
     use_color = (
         (not args.no_color) and sys.stdout.isatty() and os.environ.get("NO_COLOR") is None
     )
+    sid: str | None
     if not args.session:
+        sessions = dz.list_sessions()
         if args.json:
             rows = [
                 {
@@ -315,16 +317,33 @@ def cmd_dissect(args: argparse.Namespace) -> int:
                     "distil_input_tokens": o.distil_tokens,
                     "status": o.status,
                 }
-                for o in dz.list_sessions()
+                for o in sessions
             ]
             print(json.dumps(rows, indent=2))
+            return 0
+        print(dz.render_sessions_text(sessions, color=use_color))
+        # On a terminal, let the user pick right here instead of retyping the id.
+        if not (sessions and sys.stdin.isatty() and sys.stdout.isatty()):
+            return 0
+        try:
+            choice = input(f"\ndissect which session? [1-{len(sessions)}, Enter to quit] ").strip()
+        except (EOFError, KeyboardInterrupt):
+            print()
+            return 0
+        if not choice:
+            return 0
+        if choice.isdigit() and 1 <= int(choice) <= len(sessions):
+            sid = sessions[int(choice) - 1].sid
         else:
-            print(dz.render_sessions_text(dz.list_sessions(), color=use_color))
-        return 0
-    sid = dz.resolve_sid(args.session)
-    if sid is None:
-        print(f"no session matches {args.session!r} — run `distil dissect` to list them")
-        return 2
+            sid = dz.resolve_sid(choice)
+        if sid is None:
+            print(f"no session matches {choice!r}")
+            return 2
+    else:
+        sid = dz.resolve_sid(args.session)
+        if sid is None:
+            print(f"no session matches {args.session!r} — run `distil dissect` to list them")
+            return 2
     d = dz.dissect(sid)
     peers = dz.list_sessions()
     if args.json:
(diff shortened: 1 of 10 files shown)


dshakes:
merged in spirit and shipped in 1.15.1 — closing this as adopted.

quick note on the mechanics: rather than a straight merge I brought your changes in file-by-file so I could layer the three review fixes on top (argv → command[:1], flock on the jsonl append, and keeping the detail write synchronous). the net result is your branch, verbatim, plus those — I diffed my proxy.py against your rebased version afterward and they're identical except for exactly those fixes. you're co-author on the commit so the history keeps your name on it.

thanks for rebasing onto 1.15.0 too, that made this painless. the report's really good — the calibration-vs-billed-usage number and the "not optimized and why" section are the kind of thing I'd never have thought to surface. appreciate it.

Issue #30, opened 2026-07-21
distil default doesn't reach agents launched from IDE extensions (VSCode, Cursor, Windsurf, Antigravity)


tolgatuncoglu:
# `distil default` doesn't reach agents launched from IDE extensions (VSCode, Cursor, Windsurf, Antigravity)

## Problem

Both strategies behind `distil default` assume the wrapped agent is launched
from an interactive shell that sources an rc file:

- **Strategy A** (alias, default): `alias claude='distil wrap ... -- claude'` in
  `.zshrc`/`.bashrc`/etc.
- **Strategy B** (`--always-on`): a persistent proxy daemon, discovered via
  `export ANTHROPIC_BASE_URL=http://127.0.0.1:<port>` in the same rc file.

Neither reaches a `claude` process started by an IDE extension. Confirmed by
inspecting the process tree of the VSCode "Claude Code" extension:

```
Code Helper (Plugin)
  └─ .../extensions/anthropic.claude-code-*/resources/native-binary/claude
```

The extension host execs the native binary by absolute path — no PATH lookup,
no login/interactive shell, no rc file sourced. `distil default`'s install
silently does nothing for this session, with no error or warning.

## Scope of the gap (2x2: editor family × OS)

|                          | macOS | Linux / Windows |
|---|---|---|
| **VSCode / Cursor** (Claude-Code-extension-based — execs the real `claude` binary) | confirmed reachable — see below | reachable — see below (untested, but the mechanism is OS-agnostic) |
| **Windsurf / Antigravity** (proprietary agent runtimes, not built on the `claude` CLI, as far as I can tell) | not reachable by any mechanism this repo currently has | same |

For the top row: verified by installing the actual `anthropic.claude-code`
extension inside Cursor (not Cursor's own separate, built-in "Agents"
feature — a different, proprietary code path) and inspecting the process
tree of both:

```
Code Helper (Plugin) → .../vscode/extensions/anthropic.claude-code-2.1.215-.../native-binary/claude
Cursor Helper (Plugin): extension-host → .../cursor/extensions/anthropic.claude-code-2.1.216-.../native-binary/claude
```

Same mechanism in both: the extension host execs its own bundled copy of the
`claude` binary directly, no shell in between. Since it's the same binary
(just vendored per-editor), it reads the same `~/.claude/settings.json` on
every launch regardless of which host launched it. That file already
supports a top-level `env` object. Strategy B currently writes
[...]

PR #31, opened 2026-07-21
distil default --always-on: wire ANTHROPIC_BASE_URL into ~/.claude/settings.json

tolgatuncoglu:
Fixes (partially) #30

`distil default --always-on` currently discovers the proxy via `ANTHROPIC_BASE_URL` exported in the shell rc file. That only reaches a `claude` process that's a descendant of a shell which sourced that rc — an IDE extension execing the `claude` binary directly never sees it, and the install silently does nothing for that session.

Confirmed via process-tree inspection for both VSCode's and Cursor's Claude Code extensions:

```
Code Helper (Plugin) → .../vscode/extensions/anthropic.claude-code-2.1.215-.../native-binary/claude
Cursor Helper (Plugin): extension-host → .../cursor/extensions/anthropic.claude-code-2.1.216-.../native-binary/claude
```

No shell in between in either case — each extension host execs its own bundled copy of the same `claude` binary directly.

The `claude` binary reads `~/.claude/settings.json` on every launch regardless of how it was started, and that file already supports a top-level `env` object. This PR adds:

- `setup.wire_settings_env` / `unwire_settings_env` — same idempotent, backup-before-clobber, preserve-everything-else pattern as the existing `wire_statusline`/`unwire_statusline`, generalized to any `(env_var, value)` pair.
- Wired into `cmd_default`'s `--always-on` path (writes it, gated to `agent == "claude"` since this settings file is Claude-Code-specific) and into `--undo` / `offboard` (removes it).
- A new `--force` flag for `distil default`, mirroring `wire_statusline`'s existing conflict/`--force` behavior, for when the user already has a different `ANTHROPIC_BASE_URL` in their settings.

**What this does not solve** (see #30 for the full breakdown): Windsurf, Google Antigravity, and Cursor's own native "Agents" feature (a separate, proprietary system from the actual Claude Code extension — confirmed by inspecting its process tree too, which shows no `claude` binary at all) don't exec the `claude` CLI or read this settings file. This fix can't reach them; doing so would need either a documented per-app config surface (none found) or a TLS-intercepting proxy, which is a separate, much larger design discussion given this repo's stated zero-trust-footprint posture in `THREAT_MODEL.md`.

**Testing:** `make test` and `make lint` both green (11 new tests covering fresh-wire, idempotency, conflict+force, key preservation, non-`claude`-agent skip, and the offboard cleanup path). Not a compression-path change, so `make gate` doesn't apply per CONTRIBUTING.md.

### DIFF
--- a/distil/cli.py
+++ b/distil/cli.py
@@ -1441,6 +1441,7 @@ def cmd_onboard(args: argparse.Namespace) -> int:
                 undo=False,
                 always_on=False,
                 no_start=False,
+                force=False,
             )
         )
         print()
@@ -1495,11 +1496,14 @@ def cmd_default(args: argparse.Namespace) -> int:
     from . import onboard
     from .setup import (
         alias_body,
+        default_settings_path,
         detect_shell,
         env_body,
         remove_managed,
         service_spec,
         service_unload_cmd,
+        unwire_settings_env,
+        wire_settings_env,
         write_managed,
     )
 
@@ -1527,6 +1531,11 @@ def cmd_default(args: argparse.Namespace) -> int:
                 print(f"✓ removed proxy service {path}")
             except OSError:
                 pass
+        if agent == "claude":  # inverse of the settings.json wiring below
+            st2, msg2 = unwire_settings_env(
+                default_settings_path(), "ANTHROPIC_BASE_URL", f"http://127.0.0.1:{args.port}"
+            )
+            print(("✓ " if st2 in ("ok", "absent", "foreign") else "✗ ") + msg2)
         print(f"  open a new terminal (or: source {rc}) to finish")
         return 0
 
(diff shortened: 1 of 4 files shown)


dshakes:
Hey @tolgatuncoglu — thank you for this, genuinely. `--always-on` is something people have been asking for, and you went and built it cleanly.

I pulled it in and tested it end to end before merging. One thing I want to be upfront about: while testing I hit a sharp edge. If the `ANTHROPIC_BASE_URL` we write into `~/.claude/settings.json` ever goes stale — say distil isn't listening on that port anymore — Claude Code dies with a silent connection-refused, and it's genuinely hard to figure out why. So before merging I added a fail-loud guard in `distil doctor` that catches a dead base URL and tells you exactly what's wrong, and I isolated the settings write in the tests so the suite can't stomp a developer's real `~/.claude/settings.json`. Your feature itself is unchanged — I just wrapped it in a seatbelt.

It's shipped in **v1.26.0** (tagged just now, PyPI publish running). I also started a `CONTRIBUTORS.md`, and you're the first name on it after me.

I'm closing this PR because the work is already integrated into `main` and released — the credit is all yours. Please keep them coming, and if you ever want to take on something bigger, ping me and we'll scope it together.

Thanks again.
— Shekhar

PR #34, opened 2026-07-26
fix(a11y): restore keyboard access to navigation, sessions, and copy controls

pjdoland:
Hi! This is the first of a small series of accessibility PRs coming out of a WCAG 2.2 pass over the docs site and the dissect portal. I've kept this one scoped to the fixes where a keyboard-only or screen-reader user could not complete the action at all, not just cases where it was merely awkward. If you'd rather I bundle the remaining fixes differently (one big PR, split by page vs. by fix type, etc.) just say so and I'll adjust for the rest of the series.

No visual design changes are intended anywhere in this PR: same colors, same layout, same fonts. Everything here is about what is reachable and operable without a mouse.

## Summary

| Fix | Where | What was broken |
|---|---|---|
| Centralize the mobile sidebar toggle | all 20 docs pages, `site.js`, `site.css` | Two pages' mobile nav button was wired to a function that did not exist anywhere (`ReferenceError`); the other 18 each had their own copy-pasted version with no keyboard support at all |
| Remove dead copy buttons, harden clipboard capture | `research.html`, `deploy-security.html`, `techniques.html`, `site.js` | Nine code blocks had a copy button wired to a function that does not exist, and the real copy button on those blocks was copying the word "copy" instead of the code |
| Make session rows reachable | `distil/dissect.py` | The session picker's rows were click-only table rows with no link, so keyboard and screen-reader users could not open a session at all |
| Make mobile table scrolling reachable | `site.js`, `site.css`, `index.html` | Wide tables on mobile screens scrolled sideways, but there was nothing inside them a keyboard could focus, so that scroll was mouse (or touch) only |

### 1. Mobile sidebar navigation was either broken or keyboard-inaccessible

Every docs page has a hamburger button that is supposed to open the left-hand navigation on narrow screens. On `benchmarks.html` and `adoption.html`, that button called a JavaScript function, `toggleSidebar()`, that was never actually defined anywhere in the codebase. Clicking it did nothing but throw an error in the console: the mobile nav on those two pages was completely dead.

The other 18 pages each had their own copy-pasted version of `toggleSidebar()` sitting at the bottom of the page. It worked with a mouse, but it never told assistive technology whether the menu was open or closed, there was no way to close it with the keyboard (only by clicking outside it with a mouse), and if you did tab your way in and hit Escape, nothing happened and focus was not returned anywhere sensible.

I moved this logic into `site.js` once, so every page shares the same, tested behavior: the toggle button now reports its state (`aria-expanded`) so a screen reader announces "collapsed" or "expanded," clicking outside the open menu still closes it (same as before), and pressing Escape now closes the menu and puts keyboard focus back on the button that opened it, exactly the pattern screen reader and keyboard users expect from a menu. (WCAG references: 2.1.1 Keyboard, 4.1.2 Name/Role/Value, 2.4.3 Focus Order.)

I also made a small CSS change so that while the mobile menu is closed, its roughly 20 links are removed from the tab order (`visibility: hidden`). Previously, tabbing through a closed page on a narrow screen still stepped through all 20 hidden sidebar links before reaching the visible page content, which was disorienting and slow for keyboard users.

### 2. Dead copy buttons, and the real one was copying the wrong text

Nine code samples across three pages had a "copy" button hard-coded into the HTML, wired to a function called `copyCode()` that does not exist anywhere in this codebase. Clicking it did nothing.

Meanwhile, `site.js` already adds its own working copy button to every code block automatically, so those nine blocks ended up with two buttons stacked on top of each other: a dead one and a working one. Worse, because the working button grabs "all the text in this box" and the dead button's own label ("copy") was part of that text, clicking the button that did work copied the code with the word "copy" stuck onto it.

I removed the nine dead buttons and hardened the real one so it always reads the code cleanly, regardless of what other markup happens to be sitting in the box. (WCAG references: 3.2.4 Consistent Identification, 4.1.2 Name/Role/Value, 2.1.1 Keyboard.)

### 3. Session rows in the dissect portal had no keyboard path at all

`distil dissect` serves a small local page listing your recorded sessions so you can click into one. Each row was rendered as a plain table row with an `onclick` handler and nothing else: no link, no tab stop, no keyboard handler. If you could not or did not use a mouse, there was no way to open a session from that page, full stop.

I put a real link around the session ID in each row, so it is a normal, focusable, screen-reader-friendly link. Clicking anywhere else in the row still opens the session too, exactly as before, that convenience is untouched. (WCAG reference: 2.1.1 Keyboard.)

### 4. Wide tables on mobile scrolled, but only for a mouse or a finger

Several docs pages have data tables wider than a phone screen (comparison tables, benchmark tables). On narrow screens these already scroll sideways so you can see the rest of the columns, that part worked. The problem is there was nothing inside that scrolling area a keyboard could focus, so a keyboard-only user (this is a known Safari behavior in particular) had no way to trigger that scroll and simply could not see the columns that ran off the edge of the screen.

I wrap each table in a labeled, focusable region (you can tab to it, and its name announces which table it is, taken from the heading just above it). Scrolling now happens on that wrapper instead of the table directly, so arrow keys work once it's focused. Desktop layout is identical either way. (WCAG reference: 2.1.1 Keyboard.)

One page, `index.html`, has its own separate styling and does not use the shared page layout the other 19 pages do, so the generic fix above does not reach its one wide table. It does load `site.js` and does have a table that genuinely scrolls on narrow screens, so I applied the same fix directly on that page too, in its own inline script, rather than skip it.

## Gates

- **Lint:** `make lint` fails, but on 428 pre-existing errors unrelated to this diff (mostly `noqa` cleanup and style items across the wider codebase, none on lines this PR touches). Running `ruff check` on just `distil/dissect.py` shows its 18 pre-existing errors are all elsewhere in the file; none on the lines this PR changed.
- **Pytest:** `uv run pytest tests/test_dissect.py -x -q` passes (76 passed). Grepped `tests/` for `render_sessions_html`, `toggleSidebar`, and `copy-btn`; the one hit (`tests/test_coverage_1151.py`) only asserts the sessions HTML is truthy, so it is unaffected and still passes. Ran the full suite too: `1842 passed, 5 skipped` in about three minutes, no failures.
- **Playwright, mobile sidebar (`benchmarks.html`, `adoption.html`, `getting-started.html` at 400x800):** toggle opens/closes the sidebar, `aria-expanded` flips true/false, Escape closes it and returns focus to the toggle button, outside-click still closes it on the previously-working page, zero console errors on any of the three pages. Pass.
- **Playwright, sidebar closed at mobile width:** `getComputedStyle` on the sidebar reports `visibility: hidden` when closed, confirming its links are out of the tab order. Pass.
- **Playwright, `research.html` copy buttons:** exactly one `.copy-btn` per `<pre>` (checked all six on the page). Clicked one and read `navigator.clipboard.readText()` directly: the clipboard held the exact command text, no leftover "copy" text. Zero console errors. Pass.
- **Playwright, `cli.html` at 900x800:** all 25 tables on the page are wrapped in `.table-scroll` with `tabindex="0"`, `role="region"`, and a derived `aria-label` (e.g. "Flags"). Focused a wrapper and confirmed arrow-key scrolling moves `scrollLeft`. Pass.
- **Playwright, `index.html`:** its one wide table is wrapped the same way at 400px width (focusable, labeled, scrolls), and at 1400px the wrapper shows no overflow at all, so desktop is unchanged. Zero console errors. Pass.
- **`dissect.py` targeted check:** built a minimal `SessionOverview` and called `render_sessions_html()` directly; confirmed the output contains a real `<a href="/session/...">` around the session id and that the row's `onclick` convenience is still present. Pass.

The dev server (`python3 -m http.server`) used for the Playwright checks was stopped afterward, and no artifacts from that were left behind.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,6 +16,7 @@ This file is the record.
 
 - **pliablepixels** ([@pliablepixels](https://github.com/pliablepixels)) — per-content-type keep policy (`distil/compress/keep_policy.py`), so a digest keeps a test run's verdict instead of folding the answer and keeping the noise ([#23](https://github.com/dshakes/distil/pull/23), shipped 1.15.0); and `distil dissect` — the per-session deep-dive report, including `--serve` and agent-transcript correlation ([#27](https://github.com/dshakes/distil/pull/27), shipped 1.15.1).
 - **Tolga Tuncoglu** ([@tolgatuncoglu](https://github.com/tolgatuncoglu)) — `distil default --always-on`: persist `ANTHROPIC_BASE_URL` into `~/.claude/settings.json` so Claude Code routes through distil on every launch, including sessions started from an IDE extension rather than a terminal ([#31](https://github.com/dshakes/distil/pull/31), shipped 1.26.0).
+- **PJ Doland** ([@pjdoland](https://github.com/pjdoland)): keyboard accessibility across the docs site and dissect portal (working mobile navigation, keyboard-reachable session links, copy-button cleanup, focusable table scrolling), the first in a WCAG 2.2 series ([#34](https://github.com/dshakes/distil/pull/34)).
 
 ---
 
--- a/distil/dissect.py
+++ b/distil/dissect.py
@@ -1672,7 +1672,7 @@ def render_sessions_html(sessions: list[SessionOverview]) -> str:
     rows = (
         "".join(
             f"<tr onclick=\"location='/session/{e(o.sid)}'\">"
-            f"<td><code>{e(o.sid)}</code></td><td>{e(o.tool or '?')}</td>"
+            f"<td><code><a href=\"/session/{e(o.sid)}\">{e(o.sid)}</a></code></td><td>{e(o.tool or '?')}</td>"
             f"<td>{e(_when(o.started))}</td><td>{e(_when(o.last_ts))}</td>"
             f"<td class='r'>{o.requests}</td>"
             f"<td class='r'>{100.0 * (o.baseline_tokens - o.distil_tokens) / o.baseline_tokens if o.baseline_tokens else 0.0:.1f}%</td>"
@@ -1696,6 +1696,9 @@ th{{color:#5b6177;font-size:11px;text-transform:uppercase;letter-spacing:.07em}}
 td.r{{text-align:right;color:#5ad1c9;font-variant-numeric:tabular-nums}}
 tbody tr{{cursor:pointer}} tbody tr:hover td{{background:#10131d}}
 code{{color:#8b7bff}} .muted{{color:#5b6177}}
+code a{{color:inherit;text-decoration:underline}}
+code a:hover,code a:focus{{color:#5ad1c9}}
+code a:focus-visible{{outline:2px solid #8b7bff;outline-offset:2px;border-radius:4px}}
 .foot{{color:#5b6177;font-size:12.5px;margin-top:22px}}
 </style></head><body><div class="wrap">
 <h1>Distil <span class="g">sessions</span></h1>
--- a/docs/adapters.html
+++ b/docs/adapters.html
@@ -22,7 +22,7 @@
 <body>
 
 <header class="topbar">
-  <button class="sidebar-toggle" onclick="toggleSidebar()" aria-label="Toggle navigation">☰</button>
+  <button class="sidebar-toggle" onclick="toggleSidebar()" aria-label="Toggle navigation" aria-expanded="false" aria-controls="sidebar">☰</button>
   <a href="index.html" class="topbar-logo"><img src="assets/logo.svg" alt="" width="22" height="22" style="border-radius:6px;vertical-align:-6px;margin-right:8px"/>Dist<b>il</b></a>
   <span class="topbar-pill">compression with a quality contract</span>
   <nav class="topbar-links">
@@ -352,17 +352,6 @@ distil proxy listening on http://127.0.0.1:8788
   </main>
 </div>
 
-<script>
-function toggleSidebar() {
-  document.getElementById('sidebar').classList.toggle('open');
-}
-document.addEventListener('click', function(e) {
-  var sb = document.getElementById('sidebar');
-  if (sb.classList.contains('open') && !sb.contains(e.target) && !e.target.closest('.sidebar-toggle')) {
-    sb.classList.remove('open');
-  }
-});
-</script>
 <script src="site.js" defer></script>
 </body>
 </html>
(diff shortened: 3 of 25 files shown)


dshakes:
Merged — shipping in 1.33.0.

On your question about how to split the rest: keep doing exactly what you did. Eight focused PRs, each scoped to one failure mode with a table of what was broken, was far easier to review than one large one. I could check each claim independently, and when two of them collided during integration I knew which intent to preserve on each side. Please don't bundle.

Two things worth calling out. The `toggleSidebar` finding isn't really an accessibility fix — I checked all 21 pages, and `adoption.html` and `benchmarks.html` were the only two calling a function that nothing defined, so the mobile menu was throwing a `ReferenceError` for every visitor on those pages regardless of how they were browsing. Same story for the nine dead `copyCode` buttons. Those are plain bugs an accessibility pass happened to surface, which is the most useful kind of audit.

One integration note: #39 rewrote the same session-picker rows to poll in place. Merged naively, your `<a>` survived in the server-rendered HTML but the first poll overwrote the cell and dropped it. The row now carries both the link and the poll key, and the poll keeps the anchor's `href` in sync (9ae46d8).

Thanks for adding yourself to CONTRIBUTORS.md — I expanded the entry to cover the whole series.

dshakes:
## Validation — #34

Ran your claims as an A/B against `466455e` (pre-merge) in a real browser, clicking the actual controls.

**The broken mobile nav reproduces exactly, on exactly the two pages you named.**

| | `adoption.html` | `benchmarks.html` |
|---|---|---|
| pre-merge `typeof toggleSidebar` | `undefined` | `undefined` |
| pre-merge click result | sidebar did **not** open | sidebar did **not** open |
| pre-merge console | `Uncaught ReferenceError: toggleSidebar is not defined` | same |
| post-merge | opens, `aria-expanded` → `true` | opens, `aria-expanded` → `true` |
| post-merge console | clean | clean |

**Dead copy buttons — your count was exact.** Pre-merge, blocks wired to a `copyCode` that `typeof` reports as `undefined`: `research.html` 5, `deploy-security.html` 3, `techniques.html` 1 = **9**. Post-merge: 0.

**One claim I could not reproduce**, in fairness to the record: the secondary point that the working copy button "was copying the word *copy*". I intercepted `navigator.clipboard.writeText` on both versions; both captured the correct code (234 chars, starting `distil calibrate \`). No block's text ended in a stray "copy". Removing the dead buttons is still right — I just can't confirm that part.

**Session links** (dissect portal, live): 99 rows, **99** `code a[href^="/session/"]`. Keyboard-reachable and activatable with Enter.

**Toggle state**: `aria-expanded="false"` → `"true"` on click, `aria-controls="sidebar"` resolves. No inline `toggleSidebar` left on any of the 22 pages.

---

PR #35, opened 2026-07-26
fix(a11y): announce status changes and repair ARIA wiring in the docs

pjdoland:
## Summary

This is another installment in the accessibility remediation series that started with #34. It's a small, focused batch: two ARIA-status announcement fixes, one duplicate-id bug in the shared docs script, and one incomplete widget pattern on the landing page. Nothing here touches Python, layout, or color; it's all wiring that was either silently broken or silently missing for screen reader users.

**The fun fact worth knowing:** the duplicate-id bug (see below) means 23 of the 24 "Flags" links in `cli.html`'s page outline currently jump to the wrong section, for every single user who clicks them, sighted or not. This wasn't an assistive-tech-only bug; it's just that nobody happened to click the third "Flags" link recently.

References use WCAG 2.2 success criterion numbers (SC) as citations only; no normative claims beyond what's fixed here.

---

### Duplicate heading ids break the "On this page" TOC and anchors (SC 4.1.2)

`site.js` builds each heading's `id` by slugifying its text, with no de-duplication. `cli.html` has 24 `<h3>Flags</h3>` headings (one per CLI subcommand), so all 24 got `id="flags"`. Every "On this page" TOC entry labeled "Flags" pointed at the same element, so clicking any of them except the first one scrolled you to the wrong command's flag table. The injected "#" permalink on each heading, and the scrollspy highlighting, had the same problem.

The fix prefixes `h3` ids with their nearest `h2`'s id (e.g. `bench-flags`, `certify-flags`, `holdout-flags`), falling back to a numeric suffix for any collision that scheme doesn't resolve. Verified in-browser: all 63 ids on `cli.html` are now unique, and two different "Flags" TOC links now scroll to genuinely different points in the page (roughly 575px and 6543px down the document).

### Copy-to-clipboard confirmation was silent for screen readers (SC 4.1.3)

Every code block's Copy button swaps its own label to "Copied" for 1.6 seconds. That's a purely visual cue; there was no live region, so screen reader users who clicked Copy had no way to know the clipboard write actually happened. Added one shared, visually-hidden `aria-live="polite"` region that announces "Copied to clipboard" (or a failure message in the rare execCommand-fails fallback path). Verified the region's text updates immediately after clicking a Copy button.

### Adoption page's live status paragraph was never announced (SC 4.1.3)

`adoption.html` has a `<p id="updated">` that starts as "connecting to the metrics branch…" and gets rewritten to report success, staleness, or failure of the live data fetch. That's a textbook status message, but it had no `role="status"` or `aria-live`, so screen reader users never learned whether the numbers on the page were live, stale, or broken. Added `role="status"`, which is appropriately quiet (this text changes at most every 45 seconds). Left the fast-polling rate labels and the odometer alone, exactly as the audit recommended, since those update every few seconds and would flood a screen reader with noise if wired into a live region.

### Compression calculator results updated silently (SC 4.1.3)

`learn-compression.html` has an interactive calculator: three sliders and a model picker recompute three dollar figures on every input. The figures were plain text with no live region, so a screen reader user adjusting a slider had to manually go find the results afterward; the whole point of the widget (see the numbers move as you change the inputs) was lost. Added a debounced, visually-hidden `aria-live="polite"` summary that announces all three figures 500ms after the last input event settles, so dragging a slider doesn't fire a wall of announcements. The visible stat boxes still update instantly for sighted users; only the screen-reader summary is debounced.

### Installer tabs on the landing page were half-implemented (SC 4.1.2)

The install-format switcher (`uvx` / `pipx` / `Homebrew` / `Docker` / `zipapp`) declared `role="tablist"` and `role="tab"`, which tells a screen reader "this is a tab widget, expect selection state and arrow-key navigation." Neither existed: no `aria-selected`, no roving `tabindex`, no arrow-key handling, no `role="tabpanel"`. The widget itself worked fine by mouse and by Tab-plus-Enter, since these are native `<button>` elements; the mismatch was purely in what screen readers were told to expect versus what was actually there.

I chose to drop the tablist/tab roles and use plain buttons with `aria-pressed` instead, rather than build out the full ARIA tabs pattern (roving tabindex, ArrowLeft/ArrowRight handling, ids wiring tabs to panels). Reasoning: this is a five-way format picker wired with nothing more than a plain click handler in `site.js` (the same handler every other `.tabs` widget on the site would use, if there were others); adding a second navigation contract (arrow keys) on top of working Tab-and-Enter behavior is more surface area to maintain for no real benefit to this particular widget, and the two-line `aria-pressed` toggle change is the smaller, lower-risk fix. Verified in-browser: focusing the "pipx" button and pressing Enter toggles `aria-pressed` correctly on all five buttons and switches the visible panel, with zero console errors.

### Invalid aria-label on a generic div (SC 4.1.2)

`adoption.html` had `<div class="split" id="splitband" aria-label="real vs bot download split">`. Under ARIA 1.2, a plain `div` has an implicit role of `generic`, and generic elements aren't allowed to carry an accessible name, so browsers and assistive tech either ignore that `aria-label` or expose it inconsistently. Added `role="img"` so the element has a role that supports naming. The adjacent legend already states the real numbers, so no information depends on this label either way; it's a validity fix, not a content fix.

---

## Gates

- **No Python files changed.** This PR is docs/site.js/site markup only (`docs/site.js`, `docs/index.html`, `docs/adoption.html`, `docs/learn-compression.html`). The pytest suite is unaffected; confirmed by grepping `tests/` for any of the HTML ids/strings touched here (`role="tab"`, `copy-btn`, `te-naive`, `te-live`, `splitband`, `id="updated"`, `flags"`): the only hits are unrelated Python dict keys named `"flags"` in manifest test fixtures (`tests/test_dissect.py`, `tests/test_upgrade.py`, `tests/test_coverage_1151.py`), not the docs markup. No pytest was actually runnable in this environment (module not installed), but since no Python changed, that's moot.
- **`make lint` not relevant**, per scope (no Python changed).
- **Browser-verified** with a real Playwright session against this worktree's docs on port 8647: all 63 heading ids on `cli.html` are unique post-fix, two different "Flags" TOC links scroll to different document offsets, the copy button's live region updates after a click, the installer tabs are fully keyboard-operable (`aria-pressed` and panel state update correctly on Enter) with zero console errors on `index.html`, `cli.html`, or `adoption.html`, and the compression calculator's live region populates only after the 500ms debounce (confirmed empty at 0ms and 200ms, populated at 700ms).
- **Working tree clean**, 4 commits on `fix/a11y-aria-status`, branched from `main`.
- **Deviations / out of scope:** none. All six findings (M8, M9, M10, M11, N12, N9) were implemented as scoped; no other findings from the audit were touched, and CONTRIBUTORS.md was not modified.

### DIFF
--- a/docs/adoption.html
+++ b/docs/adoption.html
@@ -201,7 +201,7 @@
     <div class="seyebrow live gr" style="margin-top:0">Live telemetry &middot; auditable end-to-end</div>
     <h1>Adoption &amp; <span class="g">Community Savings</span></h1>
     <p class="lead">Most projects show you raw download counts and hope you don't ask questions. We ask them for you: downloads are <b>bot-filtered</b>, savings come from an <b>opt-in census</b> that starts at zero and can't be inflated, and every number below is computed by public code from a <a href="https://github.com/dshakes/distil/tree/metrics">public git branch</a>.</p>
-    <p class="updated" id="updated">connecting to the metrics branch&hellip;</p>
+    <p class="updated" id="updated" role="status">connecting to the metrics branch&hellip;</p>
 
     <!-- ── next-gen live hero: ticking odometer + trust ring ── -->
     <div class="hero">
@@ -250,7 +250,7 @@
     <div class="seyebrow gr">Downloads &middot; honestly</div>
     <h2 style="margin-top:0">The number everyone shows <span class="g">vs. the one that's true</span></h2>
     <p>A public PyPI package receives constant traffic from supply-chain scanners (Snyk/Socket-class), package-intel crawlers (deps.dev, libraries.io), and malware sandboxes — every new release triggers fleet-wide sweeps. They report no operating system; <code>pip</code> on a real machine does. Even the filtered number counts <em>events</em> (CI reinstalls, upgrades), not people — provable people are the census tile. So we split:</p>
-    <div class="split" id="splitband" aria-label="real vs bot download split">
+    <div class="split" id="splitband" role="img" aria-label="real vs bot download split">
       <div class="seg real" id="seg-real" style="flex-grow:1"></div>
       <div class="seg bots" id="seg-bots" style="flex-grow:1"></div>
     </div>
--- a/docs/index.html
+++ b/docs/index.html
@@ -611,12 +611,12 @@ VERDICT: <span class="w">FAIL</span>  (would degrade quality — blocked)</pre>
   <h2>Install <span class="g">— pick a format</span><a class="hlink" href="#install" aria-label="Link to this section">#</a></h2>
   <p class="lead">The stdlib-only core makes the packaging clean: zero runtime deps, corpus bundled.</p>
   <div class="tabs" id="install-tabs">
-    <div class="tabbar" role="tablist">
-      <button class="tab is-active" data-tab="uvx" role="tab">uvx <span class="tab-hint">zero install</span></button>
-      <button class="tab" data-tab="pipx" role="tab">pipx <span class="tab-hint">isolated cli</span></button>
-      <button class="tab" data-tab="brew" role="tab">Homebrew</button>
-      <button class="tab" data-tab="docker" role="tab">Docker</button>
-      <button class="tab" data-tab="pyz" role="tab">zipapp</button>
+    <div class="tabbar">
+      <button class="tab is-active" data-tab="uvx" aria-pressed="true">uvx <span class="tab-hint">zero install</span></button>
+      <button class="tab" data-tab="pipx" aria-pressed="false">pipx <span class="tab-hint">isolated cli</span></button>
+      <button class="tab" data-tab="brew" aria-pressed="false">Homebrew</button>
+      <button class="tab" data-tab="docker" aria-pressed="false">Docker</button>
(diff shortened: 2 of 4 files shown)


dshakes:
Merged — shipping in 1.33.0.

The live-region wiring is the part I'd have been most likely to get wrong myself. Announcing the odometer through a `role="status"` region while hiding the spinning digits is the right call, and it's a real conflict with #40, which wanted to expose that element directly. I kept your version: the digits are already announced through the status region, so exposing the raw spinner too would just double-announce.

dshakes:
## Validation — #35

Ran the live web dashboard (`distil dashboard --web`) and watched the region update.

- `role="status"` region present, `sr-only`, and **populated at runtime**: `"1.5B tokens saved, 32.2% smaller, 7,751 requests."`
- Odometer `#odo` stays `aria-hidden="true"` while showing `1,522,590,876`.

That pairing is exactly why I kept your version over #40's, which wanted to expose the odometer element directly. The digits are already announced as a sentence; exposing the spinner too would have made a screen reader read the raw number churn on every tick. Your split — human summary to AT, animation to the eye — is the correct one.

---

PR #37, opened 2026-07-26
fix(a11y): expose diagrams, charts, and their links to assistive technology

pjdoland:
Hi team, this is the next installment in the accessibility remediation series that started with #34. It closes out the "non-text content" findings from the audit: images, charts, and an interactive diagram that either say nothing to a screen reader or say the wrong thing.

## The headline fix: the architecture diagram was inviting clicks nobody could hear

`architecture.html` and `concepts.html` both ship the same interactive SVG: six pipeline stages and four "quality contract" cards, each one a real link into `techniques.html` or `concepts.html`. The page literally tells sighted visitors "Interactive: click any block to open the concept behind it." The problem: the `<svg>` had `role="img"`, which tells assistive tech "this is one flat picture, nothing inside it matters." Screen reader users got a single unlabeled image announcement and never learned the ten links existed, even though they were still sitting in the keyboard tab order, silent and unlabeled, when focused.

Fix: the svg is now `role="group"` (keeps the same overall `aria-label` describing the whole diagram), and every one of the ten links gets its own `aria-label` so its name is announced reliably (the existing `<title>` elements weren't consistently exposed cross-browser). The "click any block" copy is now "select any block (click, or Tab then Enter)" so it doesn't assume a mouse. Same change, same wording, on both pages, since they share the identical figure. (WCAG 1.1.1, 4.1.2)

## Dissect report charts now have names, and the tools chart got its missing table

`distil dissect --html` renders three charts: the per-request timeline, the tool-cost bars, and the folded-blocks bars. All three were `<svg role="img">` with nothing to tell a screen reader what they showed. They now carry a descriptive `aria-label` (and a matching `<svg><title>` as a belt-and-suspenders backup) built from the actual data, e.g. "Tokens folded by block kind, top 10 kinds." The timeline chart already had a collapsible data-table fallback; the tool-cost chart didn't, so it's the odd one out no longer, it gets the same pattern (tool name, tokens per request, session total). (WCAG 1.1.1)

## Four diagram alts on the marketing and docs pages actually describe the diagram now

`index.html`, `corpus.html`, and `benchmark.html` had a handful of images whose alt text named the topic but not the content, things like `alt="architecture"` or `alt="one proxy, every SDK"`. I opened each SVG and read what it actually draws, then wrote alts that carry the real message: the architecture alt now names the five pipeline stages and the recovery loop, the cache-aware chart alt states all four bar values, the "7 domains" chart alt lists all seven domains and their pass rates, and the cross-SDK alt names the five client SDKs and three provider groups it connects. Where a table right below the image already had the numbers (corpus.html, benchmark.html), the alt says so explicitly instead of repeating every digit. (WCAG 1.1.1)

## The adoption page's sparkline can now say what it's showing

The tiny JS-drawn sparkline on `adoption.html` (community tokens saved over time) was an `<svg>` with zero role or name. It now gets `role="img"` plus a data-driven `aria-label` computed at render time, e.g. "Community tokens saved trending from 1.2k to 8.4k over 12 censuses," so the trend it's drawing is actually stated, not just implied by a squiggly line.

## Decorative emoji stop announcing themselves

The two hero door icons and the three plain-English card headings on `index.html` use emoji purely for visual flavor, the text next to them already carries the meaning. Screen readers were reading them out by their Unicode names ("high voltage," "card file box"). They're now wrapped in `aria-hidden="true"` so they're silent, decorative flourishes the way they were meant to be. (best practice, WCAG 1.1.1 framing)

## Visual changes

None. Every fix here is either an ARIA attribute, an alt/aria-label string, or a wording tweak in visible instructional text (the "select any block" hint); no layout, color, or markup structure changed. Screenshots confirmed the architecture diagram renders pixel-identical.

## Gates

- `uv run ruff check` on every `.py` file touched (`distil/dissect.py`): pass, no new errors (18 pre-existing errors before and after, same categories).
- `uv run pytest -q`: pass, 1842 passed, 5 skipped, in 169.87s. No test assertions needed updating; the existing dissect chart tests check for content by substring (`"data table" in page`, `page.count("<svg") == 3`), which still hold.
- Browser verification (shared Playwright instance, served this worktree's `docs/` on port 8646): pass. On `architecture.html`, an accessibility snapshot shows all 10 diagram links exposed by name (e.g. "Cache-aware placement, pin the stable prefix at the 0.1x cache-read rate" and "Learned keep-model, the self-improving, never-regressing flywheel") inside a `group`, not swallowed by a flat image label; a DOM check confirms all 10 are natively focusable (`tabIndex: 0`). A screenshot of the figure matches the pre-change layout pixel for pixel, confirming no visual change. On `adoption.html`, the live sparkline's rendered aria-label read "Community tokens saved trending from 1.2B to 1.5B over 10 censuses," confirming the data-driven label works against real fetched data, not just the synthetic case.
- Dissect chart rendering: rendered a synthetic session report via the existing test-suite fixture (`tests/test_dissect.py::_seed_state`) and confirmed all 3 chart svgs (timeline, tools, block kinds) carry both `role="img"` and a descriptive `aria-label`, and the tools chart now has its own collapsible data table alongside the existing timeline one (2 "data table" fallbacks total, up from 1).

### DIFF
--- a/distil/dissect.py
+++ b/distil/dissect.py
@@ -1076,11 +1076,15 @@ def _svg_hbars(
     rows: list[tuple[str, int, list[tuple[str, str, str]]]],
     *,
     color: str = _C_OVERHEAD,
+    name: str = "Bar chart",
 ) -> str:
     """Horizontal bar chart (single measure, one hue): label, thin bar, value.
 
     Each row is (label, value, tooltip_rows). The whole row — label through
-    value — is one oversized hit target for the styled tooltip.
+    value — is one oversized hit target for the styled tooltip. ``name`` is
+    the chart's accessible name (SC 1.1.1): it is rendered both as the
+    ``aria-label`` and as an ``<svg><title>`` so it survives in AT that
+    ignores one or the other.
     """
     if not rows:
         return ""
@@ -1089,7 +1093,9 @@ def _svg_hbars(
     plot_w = width - label_w - val_w
     parts = [
         f'<svg viewBox="0 0 {width} {len(rows) * rh + 6}" role="img" '
+        f'aria-label="{_html.escape(name, quote=True)}" '
         f'style="width:100%;height:auto;font:12px Inter,ui-sans-serif,sans-serif">'
+        f"<title>{_html.escape(name)}</title>"
     ]
     for i, (label, val, tip_rows) in enumerate(rows):
         y = i * rh + 4
@@ -1129,9 +1135,16 @@ def _svg_stack_timeline(requests: list[dict[str, Any]]) -> str:
     n = len(comp)
     step = plot_w / n
     bar_w = max(2, min(28, round(step - 2)))
+    total_saved = sum(saved for _o, _s, saved, _r in comp)
+    name = (
+        f"Per-request token composition for {n} requests: stacked overhead, sent content, "
+        f"and tokens saved, {_human(total_saved)} saved in total"
+    )
     parts = [
         f'<svg viewBox="0 0 {width} {height}" role="img" '
+        f'aria-label="{_html.escape(name, quote=True)}" '
         f'style="width:100%;height:auto;font:11px Inter,ui-sans-serif,sans-serif">'
+        f"<title>{_html.escape(name)}</title>"
     ]
     for frac in (0.0, 0.5, 1.0):  # recessive grid, three lines
         y = 8 + plot_h * (1 - frac)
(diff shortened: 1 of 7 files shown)


dshakes:
Merged — shipping in 1.33.0.

The data-table fallbacks matter more here than the alt text. The dissect charts are how people check where their tokens went, and before this the only way to get an exact number off a bar was to hover it — which also meant no exact numbers for anyone reading the report in a terminal browser, a screenshot, or a printout. Same values, now in a table.

dshakes:
## Validation — #37

Checked against a real dissect report (243-request session).

- **2 of 2** `svg[role="img"]` carry both `aria-label` **and** a child `<title>` — belt and braces, as your docstring says.
- The generated name is genuinely descriptive, not a label of convenience: `"Per-request token composition for 243 requests: stacked overhead, sent content, and tokens saved, …"`
- **2** `data table` disclosures present, carrying the same numbers as the bars.

The fallback tables are the part I think matters most here. Those charts are how someone checks where their tokens went, and hovering a bar was the only way to get an exact figure — which also meant no exact figures in a screenshot, a printout, or a terminal browser.

---

PR #38, opened 2026-07-26
fix(a11y): visible focus, skip links, and an accessible dissect tooltip system

pjdoland:
Hey team. This is the next installment in the accessibility remediation series that started with #34: it follows on from the keyboard, ARIA, contrast, and text-alternative fixes and covers navigation and focus. Nothing here changes how the site looks, except that a couple of controls now show a visible focus outline where they didn't before (that's the point).

## Skip link on every page

Before: on every one of the 21 docs pages, a keyboard user had to tab through the top bar and the roughly 20-link sidebar (about 24 stops) before reaching the actual page content. There was no way to skip past it.

After: pressing Tab once now surfaces a "Skip to main content" link at the very top of every page. Activating it jumps straight to the content area. It's invisible until it receives keyboard focus, so sighted mouse users won't notice any change.

The landing page (index.html) doesn't have a `<main>` element yet, so its skip link points at the hero section instead, its first real content block today. A follow-up PR is adding a proper `<main>` there.

WCAG reference: 2.4.1 (Bypass Blocks).

## Copy button visible when tabbed to

Before: every code block on every docs page has a "Copy" button that only appeared on mouse hover. A keyboard user tabbing to it landed on an invisible button with no visible focus ring at all, there was simply no way to tell it was there.

After: the button now appears as soon as it receives keyboard focus, same as it does on hover.

WCAG reference: 2.4.7 (Focus Visible).

## "On this page" outline in the right place

Before: the little "On this page" navigation box floats visually at the top-right of every docs article, but in the underlying document it was tacked on after the site footer. That meant screen reader users and keyboard users only reached it after reading (or tabbing through) the entire page, the opposite of where it visually sits.

After: it's now inserted at the top of the article in the document itself. Nothing changes visually, it's still the same floating box in the same place, but reading order and tab order now match what you see.

WCAG reference: 1.3.2 (Meaningful Sequence), 2.4.3 (Focus Order).

## Heading links that don't repeat themselves

Before: every section heading has a small "#" link for deep-linking, and every single one of them had the exact same accessible label: "Link to this section." A screen reader user jumping through the page's heading list heard dozens of identically named links and no way to tell them apart.

After: each one now announces the section it actually links to, e.g. "Link to section: Install."

WCAG reference: 2.4.6 (Headings and Labels).

## Dissect portal 404 pages

Before: when the `distil dissect` session portal returned a "not found" response, it sent a bare HTML fragment: no title, no doctype, no declared language. Browsers and assistive tech had nothing to announce for the page.

After: both 404 responses are now complete, minimal HTML documents with a proper title and `lang="en"`. Same wording, same link back to the session list.

WCAG reference: 2.4.2 (Page Titled), 3.1.1 (Language of Page).

## Dissect report tooltips, now genuinely usable

This was the biggest chunk of the PR: the dissect report page has stat tiles, chart marks, and table headers that show a custom-styled tooltip on hover or focus. Four things were wrong with it:

- **No focus indicator at all.** The tiles were keyboard-focusable but the stylesheet explicitly suppressed the outline (`outline: none`), so a keyboard user tabbing through the page had no idea where they were. Now there's a clear purple focus ring.
- **The tooltip wasn't wired to its trigger.** A screen reader user could focus a tile and hear nothing about what its tooltip said. Each trigger now points at the tooltip via `aria-describedby`, so its content gets announced.
- **You couldn't dismiss or hover into the tooltip.** There was no Escape key to close it, and it vanished the instant your mouse left the trigger, so you could never read a long tooltip by hovering into it. Escape now closes it, and you can now move the mouse onto the tooltip itself to read it.
- **Some explanations only existed in a mouse tooltip attribute.** The three headline stat cards, and the "seen"/"restore" column headers, explained themselves only through a browser `title` attribute, invisible to keyboard and touch users. They now go through the same accessible tooltip mechanism as everything else on the page.

WCAG reference: 2.4.7 (Focus Visible), 4.1.2 (Name, Role, Value), 1.4.13 (Content on Hover or Focus), 1.1.1 (Non-text Content).

## Gates

- `uv run ruff check` on the one Python file touched (`distil/dissect.py`): 18 pre-existing errors before and after my changes, no new ones introduced. (Repo-wide `ruff check` still fails with the ~428 pre-existing errors noted in the audit; unrelated to this PR.)
- `uv run pytest -q`: 1842 passed, 5 skipped, 0 failed, both before committing and again on the final branch state.
- Grepped `tests/` for every string this PR touches (`data-tip`, `outline`, the 404 fragment text, `hanchor`): the one hit (`tests/test_dissect.py`'s `data-tip=` count assertion) is a `>=` lower bound that my additions only push higher, so it still passes; confirmed by running that test file directly.
- Browser-verified against a live server on the shared Playwright browser (with the lock properly acquired and released): skip link is the first Tab stop and moving focus into main/header on activation, keyboard focus reveals the copy button (`opacity: 1`), the TOC now precedes the footer in DOM order, heading anchor labels are unique, and on a synthetically rendered dissect report: focus-visible outline shows on a tooltip tile, `aria-describedby` resolves to the tooltip element, the tooltip's content is announced, Escape hides it, and the pointer can move onto the tooltip without it disappearing.

No deviations from the plan. Everything in scope landed; nothing was cut.

### DIFF
--- a/distil/dissect.py
+++ b/distil/dissect.py
@@ -1224,6 +1224,10 @@ def _tip_attr(title: str, rows: list[tuple[str, str, str]] | None = None, body: 
     innerHTML), so model/tool/signature names stay data, not markup. ``rows``
     are (series_color, label, value) triples; values render strong, labels
     secondary, keyed by a short stroke of the series color.
+
+    Also wires ``aria-describedby="tip"`` so a screen reader announces the
+    same explanation a sighted mouse or keyboard user sees in the floating
+    tooltip (the shared ``#tip`` element rendered once per page).
     """
     payload: dict[str, Any] = {"t": title}
     if rows:
@@ -1232,7 +1236,7 @@ def _tip_attr(title: str, rows: list[tuple[str, str, str]] | None = None, body: 
         payload["body"] = body
     return (
         f' data-tip="{_html.escape(json.dumps(payload, ensure_ascii=False), quote=True)}"'
-        ' tabindex="0"'
+        ' tabindex="0" aria-describedby="tip"'
     )
 
 
@@ -1481,7 +1485,7 @@ text. Only these labels are stored, never the content.</p>
 <p class="desc">The biggest single blocks distil summarized. The handle is the short ID the
 model can use to ask for the original back; “recoverable” means those original bytes are
 still on this machine.</p>
-<table><tr><th>handle</th><th>kind</th><th>tokens</th><th title="how many requests carried this block">seen</th><th title="is the original still on disk (restore/)?">restore</th></tr>{top_rows}</table>"""
+<table><tr><th>handle</th><th>kind</th><th>tokens</th><th{_tip_attr("seen", body="How many requests carried this block.")}>seen</th><th{_tip_attr("restore", body="Is the original still on disk (restore/)?")}>restore</th></tr>{top_rows}</table>"""
     else:
         detail_body = (
             "<h2>Request detail</h2><p class='muted'>Not recorded — per-request detail needs a "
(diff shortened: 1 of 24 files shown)


dshakes:
Merged — shipping in 1.33.0.

Dropping `pointer-events: none` from the tooltip is the change I looked at hardest, since a tooltip that can swallow clicks would be a bad trade. It's fine: `place()` offsets the panel 14px from the cursor and keeps that offset when it flips, so it never sits under the pointer, and your `mousemove` guard handles the hover-onto-tip case. Escape-to-dismiss and the focus-visible outlines are straightforwardly right.

The complete 404 documents are a small thing that I appreciated — `<h1>not found</h1>` with no document shell was sloppy of me.

dshakes:
## Validation — #38

**Skip link**, driven by real `Tab`/`Enter`: first tab stop on the page, `top` moves `-59px` → `0px` on focus, and Enter lands focus on `MAIN#content`.

**Tooltip**, on a real dissect report — WCAG 1.4.13 on all three sub-criteria:

| behaviour | result |
|---|---|
| hidden initially | `display: none` |
| appears on **focus** (not just hover) | `display: block`, correct content |
| dismissible | `Escape` → `display: none` ✅ |
| hoverable | `pointer-events: auto` ✅ |

**267 of 267** `[data-tip]` elements are `tabindex="0"` and `aria-describedby="tip"`; `#tip` resolves.

**On the `pointer-events` risk I raised in review** — I measured it rather than trusting the code. With the cursor at `(478, 360)`, the tooltip box starts at `x=492`: a **14px gap**, `cursorInsideTip: false`, and `elementFromPoint` returns the card underneath, not the tip. It cannot swallow a click. Your `place()` offset holds even when the panel flips.

---

PR #39, opened 2026-07-26
fix(a11y): give users control over timed refreshes and auto-updating content

pjdoland:
This is the next installment in the accessibility series that started with #34. Where that one was about keyboard reachability and broken controls, this one is about time: several pages either reload themselves on a clock or update their content continuously, and until now there was no way to pause, stop, or extend any of it. That's a real problem for keyboard and screen reader users (their reading position gets yanked out from under them), and it's also just not great for anyone reading closely on a slow connection or a distracting monitor.

Nothing here changes what data is shown or how often it refreshes by default. It only adds a way to stop the clock when you want to.

## Gateway dashboard (`distil/gateway.py`)

**Before:** the dashboard reloaded the entire page every 5 seconds via a `<meta http-equiv="refresh">` tag. Full reloads reset scroll position, keyboard focus, and a screen reader's place in the table every single cycle, with no way to turn it off.

**Now:** the page polls the existing `/distil/stats` JSON endpoint and patches the headline cards and tenant table in place, no reload. A visible Pause/Resume button next to the refresh note lets you freeze it entirely. Same data, same 5-second default cadence, just no more forced reloads.

(WCAG 2.2.1 Timing Adjustable)

## Sessions portal (`distil/dissect.py`)

**Before:** same story, 15-second full-page reloads via meta refresh. On a portal whose whole job is letting you scan and click into a session, getting reloaded before you finish reading was a real problem.

**Now:** added a small `/sessions.json` endpoint and the page polls it every 15 seconds, updating existing table rows by session id instead of tearing down and rebuilding the table. I verified in a real browser that the same row element survives an automatic poll cycle, which is what keeps keyboard focus and reading position intact once a future PR makes rows keyboard-focusable. A Pause/Resume button is there too.

(WCAG 2.2.1 Timing Adjustable)

## Live savings dashboard (`distil/webdash.py`)

**Before:** this page polls every second and runs a permanent animation loop for the odometer, with no pause control anywhere, and no screen reader ever finds out the numbers changed, since nothing announced updates.

**Now:** a Pause/Resume button stops the poll, cancels the animation loop, and pauses the pulsing status dot. Separately, added a quiet status region that composes a plain sentence ("12,345 tokens saved, 40% smaller, 12 requests") and updates at most once every 30 seconds, so screen reader users get an occasional honest summary instead of either silence or a second-by-second flood. The fast-ticking odometer digits themselves are marked so they're not read aloud digit by digit.

(WCAG 2.2.2 Pause, Stop, Hide; WCAG 4.1.3 Status Messages)

## Adoption page (`docs/adoption.html`)

**Before:** polls a live endpoint every 3 seconds and the metrics branch every 45 seconds, with a permanent animation loop for its own odometer, no pause control.

**Now:** a "Pause live updates" button next to the odometer stops both polling intervals and the animation loop; clicking it again resumes everything immediately. Also had to fix a small bug this surfaced: the "live/exact" label was being rewritten with `innerHTML` in a way that would have wiped out the new button, so that update now targets a small child span instead.

(WCAG 2.2.2 Pause, Stop, Hide)

## Decorative motion (index.html, site.css, adoption.html)

Several purely decorative animations, the landing page's hero illustration, the blinking terminal cursor, the architecture diagram's flowing lines, and a few pulsing dots on the adoption page, played on an infinite loop with no way to stop them. All of these already respect `prefers-reduced-motion`, but that's an OS-level setting, not a page control, so anyone who finds the motion distracting without having set that preference had no recourse.

Rather than add another toggle, I capped each animation's iteration count so it plays for a few seconds and then settles, instead of running forever. Visually nothing changes except that the motion stops repeating after it's made its point.

(WCAG 2.2.2 Pause, Stop, Hide, a mitigated finding, not a hard failure, since reduced-motion already covers the strict case)

## Reduced-motion smooth scroll (site.css, index.html)

Small but worth calling out: the rule meant to turn off smooth scrolling for `prefers-reduced-motion` users was written as a universal selector (`*, *::before, *::after`), which loses a CSS specificity fight against the plain `html { scroll-behavior: smooth }` rule it was supposed to override. So smooth scrolling silently stayed on for exactly the users who'd asked for it to be off. Added a same-specificity `html` rule inside the existing reduced-motion media query so the override actually wins.

(WCAG 2.3.3 Animation from Interactions)

## Gates

- `uv run ruff check` on every Python file touched (`distil/gateway.py`, `distil/dissect.py`, `distil/webdash.py`): compared error counts against the unmodified file in place (not a copy elsewhere, since ruff's config resolution depends on being inside the repo). Zero new errors in any of the three files.
- `uv run pytest -q`: full suite passes, 1844 passed, 3 skipped (skips are pre-existing and unrelated), twice, once before the docs/CSS commits and once after the full branch was assembled.
- Grepped `tests/` for `meta http-equiv`, `refresh-note`, `poll`, `render_sessions_html`, and related strings. Only one test asserted the old meta-refresh markup (`test_dashboard_html_contains_dark_bg` in `tests/test_gateway.py`); updated it to assert the meta refresh is gone and the new polling/pause markup is present, without weakening the check.
- Browser verification (Playwright, shared instance, coordinated with the other parallel agents via a lock file):
  - Gateway dashboard: rendered with synthetic tenant data, served locally. Confirmed no `meta[http-equiv=refresh]`, zero console errors, and the Pause button toggles `aria-pressed` and the visible label, and the note text, correctly in both directions.
  - Sessions portal: rendered with two synthetic sessions, served locally with a working `/sessions.json`. Confirmed no meta refresh, zero console errors, rows keyed by `data-sid`, and, ran the server for a real 15-second poll cycle and confirmed the same table row DOM node was reused afterward (not rebuilt), which is what will keep focus and reading position stable once row links land in a later PR.
  - Webdash: served the real `build_server()` handler. Confirmed the odometer is `aria-hidden`, the `role="status"` region exists and holds a composed sentence, the Pause button toggles a `body.paused` class, and, by spying on `window.fetch`, confirmed zero poll requests fire while paused and polling resumes immediately on Resume.
  - Adoption page: served the real file. Confirmed the pause button toggles state correctly and, via the same fetch-spy technique, confirmed zero requests during a 3.2-second paused window versus multiple after resuming. One unrelated console error appeared ("Unexpected token '<'") from the page's `fetchAggregates()` hitting the real GitHub API from inside the test sandbox; it's inside code this PR never touches (the JSON-parsing fallback chain), and the "live" pulse label had in fact already updated successfully once before that, so it looks like an intermittent GitHub API response shape, not a regression.
  - Reduced-motion scroll override and the capped decorative animations: confirmed statically (grep/computed-style check) that the `html { scroll-behavior: auto }` rule exists in both files and that `.hero-lab .dot`'s computed `animation-iteration-count` is now `4`, not `infinite`.

## Deviations / out of scope

- C4 (session rows are `onclick`-only `<tr>`s with no keyboard path) is a separate finding in the audit and out of scope here; the sessions-portal rework was written so that once rows become real links, focus will naturally survive the row-diffing update, but that change itself belongs to whichever PR owns C4.
- Contrast fixes (the refresh-note color, `--dim`, etc.) are explicitly a sibling PR's responsibility; I only changed the refresh-note's text, never its color.
- M11 (adoption's `#updated` status text) and S13/S14/S15 (dissect tooltip accessibility) are separate findings and untouched here.

### DIFF
--- a/distil/dissect.py
+++ b/distil/dissect.py
@@ -1666,29 +1666,58 @@ shadow.jsonl on this machine. Content-free — handles and kind:size signatures 
 
 
 # ------------------------------------------------------------------- portal
+def _session_row(o: SessionOverview) -> dict[str, Any]:
+    """One session as JSON-serializable data, shared by the initial HTML
+    render and the ``/sessions.json`` poll endpoint so both stay in sync."""
+    pct = (
+        100.0 * (o.baseline_tokens - o.distil_tokens) / o.baseline_tokens
+        if o.baseline_tokens
+        else 0.0
+    )
+    return {
+        "sid": o.sid,
+        "tool": o.tool or "?",
+        "started": _when(o.started),
+        "last": _when(o.last_ts),
+        "requests": o.requests,
+        "pct": round(pct, 1),
+        "status": o.status,
+    }
+
+
+def sessions_json(sessions: list[SessionOverview]) -> str:
+    """JSON payload backing the portal's in-place poll (``/sessions.json``)."""
+    return json.dumps([_session_row(o) for o in sessions])
+
+
 def render_sessions_html(sessions: list[SessionOverview]) -> str:
-    """The portal index: the session picker as a clickable page."""
+    """The portal index: the session picker as a clickable page.
+
+    Polls ``/sessions.json`` every 15 s and patches the table body in place
+    (keyed by session id) rather than reloading the whole page, so keyboard
+    focus and reading position survive an update. A visible Pause button
+    lets a user stop the polling entirely.
+    """
     e = _html.escape
     rows = (
         "".join(
-            f"<tr onclick=\"location='/session/{e(o.sid)}'\">"
-            f"<td><code>{e(o.sid)}</code></td><td>{e(o.tool or '?')}</td>"
-            f"<td>{e(_when(o.started))}</td><td>{e(_when(o.last_ts))}</td>"
-            f"<td class='r'>{o.requests}</td>"
-            f"<td class='r'>{100.0 * (o.baseline_tokens - o.distil_tokens) / o.baseline_tokens if o.baseline_tokens else 0.0:.1f}%</td>"
-            f"<td>{e(o.status)}</td></tr>"
-            for o in sessions
+            f"<tr data-sid=\"{e(r['sid'])}\" onclick=\"location='/session/{e(r['sid'])}'\">"
+            f"<td><code>{e(r['sid'])}</code></td><td>{e(r['tool'])}</td>"
+            f"<td>{e(r['started'])}</td><td>{e(r['last'])}</td>"
+            f"<td class='r'>{r['requests']}</td>"
+            f"<td class='r'>{r['pct']:.1f}%</td>"
+            f"<td>{e(r['status'])}</td></tr>"
+            for r in (_session_row(o) for o in sessions)
         )
         or "<tr><td class='muted' colspan='7'>no sessions recorded yet — run a wrap first</td></tr>"
     )
     return f"""<!doctype html><html lang="en"><head><meta charset="utf-8"/>
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
-<meta http-equiv="refresh" content="15"/>
 <title>Distil — sessions</title><style>
 body{{margin:0;background:#06070b;color:#f2f3f7;font:15px/1.6 Inter,ui-sans-serif,sans-serif}}
 .wrap{{max-width:820px;margin:0 auto;padding:48px 24px}}
 h1{{font-size:30px;font-weight:800;letter-spacing:-.02em;margin:0 0 6px}}
-.sub{{color:#9aa1b3;margin:0 0 28px}}
+.sub{{color:#9aa1b3;margin:0 0 28px;display:flex;align-items:center;gap:12px;flex-wrap:wrap}}
 .g{{background:linear-gradient(135deg,#8b7bff,#5ad1c9);-webkit-background-clip:text;background-clip:text;color:transparent}}
 table{{width:100%;border-collapse:collapse;border:1px solid #1b2030;border-radius:12px;overflow:hidden}}
 th,td{{padding:11px 14px;border-bottom:1px solid #1b2030;text-align:left}}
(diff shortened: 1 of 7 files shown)


dshakes:
Merged — shipping in 1.33.0.

The meta refreshes were worse than a conformance issue. The gateway dashboard reloaded itself every 5 seconds, so anyone actually watching it during a load test lost scroll position and text selection continuously; the sessions portal did the same on 15s. Polling and patching in place, with a visible pause, is better for everyone looking at those pages.

Two adjustments while integrating:

- The poll located rows by interpolating the session id into a CSS selector. Session ids come from `$DISTIL_SESSION`, so an id containing a quote would make `querySelector` throw inside the `forEach` and silently kill the loop — the page would look live and quietly stop updating. Swapped to an indexed lookup keyed off `data-sid` (null-prototype, so an id of `__proto__` is just a key).
- Merged with #34 so the row keeps its real link, and the poll now updates the anchor's `href` rather than replacing the cell text.

I also added coverage for `/sessions.json` — the payload shape, and an assertion that it stays content-free, since it's a new endpoint on the local server.

dshakes:
## Validation — #39

Tested the behaviour, not just the markup.

**Sessions portal — a live 15s poll cycle with focus parked on a link:**

| | before poll | after 18s |
|---|---|---|
| row present | ✅ | ✅ |
| anchor `href` | `/session/s1784936222-54507` | **unchanged** |
| link text | preserved | **preserved** |
| `scrollY` | 96.5 | **96.5** |
| `document.activeElement` | the anchor | **still the anchor** |

That last row is the whole point, and it's also the case that would have silently broken if #34 and this had been merged naively — the poll would have rebuilt the cell and dropped the link.

**Pause**, with `fetch` instrumented: click → label `Resume`, `aria-pressed="true"`, footnote *"Updates paused."* — **0 fetches over 19s** (>1 cycle). Resume → 1 immediate fetch, `aria-pressed="false"`, footnote back to *"Updates every 15 s."*

**Gateway dashboard**: text selection held across **16s = 3 poll cycles**. Under the old 5s `<meta refresh>` it was destroyed three times in that window. Neither surface has a meta refresh left.

`/sessions.json` returns session metadata only — no prompt or response content. Covered by a new test.

---

PR #40, opened 2026-07-26
fix(a11y): real navigation, table, and page structure for assistive technology

pjdoland:
Hey team. This is the next installment in the accessibility series that started with #34. Like the earlier PRs, it's about structure, not looks: nothing here should look any different to a sighted user. If you spot a pixel out of place, that's a bug, please flag it.

## The why, in plain terms

Sighted users navigate a page by its visual layout: a sidebar on the left, section headings in bold, a table row lining up under its column header. Screen reader users get none of that for free. They rely on the page's underlying HTML to tell them "this is navigation," "this is a heading," "this cell belongs to this row and this column." When that structure is missing, the page still *looks* fine, but it becomes a maze for anyone not looking at it.

Two concrete examples from this codebase before this PR:

- The docs sidebar (all 20 pages) was just a box of links with no `<nav>` element and no lists. A screen reader announced it as generic "complementary" content, the same way it would announce a random sidebar ad. There was no way to jump straight to it, and the "Getting started / Learn / Reference / More" grouping you can see at a glance was invisible.
- The capability matrix on the Compare page, the flag tables in the CLI reference, and every table our Python modules generate (session reports, the gateway dashboard, the savings ledger) marked every cell as generic data. A screen reader landing on a checkmark in the "rtk" column had no idea it was answering "does rtk support per-step certificates" versus any other row. It just said "checkmark."

This PR fixes both of those, plus a few smaller structural gaps in the same family: two pages had heading levels that jumped from h2 straight to h4 (screen readers use headings the way sighted users use font size to skim a page), the landing page had no `<main>` landmark for "skip to the content," and the live savings dashboard (`webdash.py`) had no heading at all and no link between a metric's number and its label.

## What changed

1. **Docs sidebar** (all 20 pages that have it): the `<aside class="sidebar" id="sidebar">` wrapper is untouched (the mobile toggle and shared CSS depend on that exact id and class), but its contents are now `<nav aria-label="Documentation">` with real `<ul>`/`<li>` lists, and the section labels ("Getting started," "Learn," etc.) are headings instead of styled divs. The active page's link gets `aria-current="page"`.
2. **Table headers**: every matrix-style table (Compare's two comparison tables, the landing page's capabilities table, Benchmark's three standings tables, and all 25 flag-reference tables in the CLI docs) now marks its row label as `<th scope="row">` and its column headers as `<th scope="col">`. A small CSS rule makes sure the row-header cell still looks exactly like the plain cell it replaced, not like a bold little column heading.
3. **Generated report tables** (`dissect.py`, `gateway.py`, `telemetry.py`, `benchmark.py`, `ledger.py`): every table gets `scope="col"` headers and a short caption naming the table (visually hidden, since a visible one would change the design). The gateway dashboard's TOTAL row moves from a bold row in the middle of the data into a proper `<tfoot>` with `th scope="row"`.
4. **Heading levels**: `deploy-security.html`, `techniques.html`, and the landing page had teaser cards using `<h4>` directly under an `<h2>`, skipping `<h3>` entirely. Promoted to `<h3>` with a matching CSS rule so they render identically.
5. **Landmarks**: the landing page's main content is now wrapped in `<main>`, matching every other docs page.
6. **webdash.py**: the page label is now a real `<h1>`, the card is wrapped in `<main>`, and each metric tile's value is linked to its label with `aria-labelledby`.

## Zero visual changes

This is the part I want to be most upfront about, since it's easy to say and harder to prove. For every markup change here, I added or adjusted CSS so nothing shifts: the sidebar's link groups are `list-style: none` with the same padding and margins as before, promoted headings get font-family/size/weight overrides so they render identically to the old tags, and the new table row-headers get their column-header styling stripped back down to plain-cell styling. I checked this in a real browser (screenshots, computed styles at both desktop and mobile widths) rather than just eyeballing the diff, and I caught and fixed one real bug in the process: my first pass hardcoded a font size on the new row-header cells that didn't track the mobile breakpoint's smaller table font, which would have made row labels look slightly larger than the data next to them on narrow screens. Fixed before landing.

One small, known side effect: promoting those teaser-card headings to real `<h3>` means the shared "on this page" table-of-contents script now picks them up (it only scans `h2`/`h3`), so on `deploy-security.html` the TOC gets a duplicate "Local dev sidecar" entry (the teaser card and the full section below share that title). This is a pre-existing site-wide issue with duplicate heading IDs (the TOC script doesn't de-duplicate slugs yet, cli.html already has this same problem with 24 "Flags" headings): not something this PR introduces, but worth knowing it now touches one more page. (#35 in this series fixes the underlying slug de-duplication, so once both land, the two entries at least get unique ids and correct scroll targets.)

## Gates

- **Lint** (`uvx ruff check distil tests`): 428 errors both before and after this branch. All pre-existing; nothing new. I ran `ruff check` on every Python file I touched individually and confirmed the count matches its own baseline.
- **Tests** (`uv run pytest -q`): 1844 passed, 3 skipped, both before my changes and after. No test assertions on table markup existed to update; I grepped `tests/` for `th`, `caption`, and `scope=` first to confirm.
- **Browser verification**: held the shared Playwright lock, served my worktree's `docs/` on port 8648, and checked: sidebar renders pixel-identical (screenshot + computed styles for list-style, margins, font-family, the accent-bar pseudo-element), the `nav` landmark and `aria-current="page"` are present in the accessibility tree, the mobile sidebar toggle still opens/closes correctly, `compare.html`'s row headers announce as `rowheader`/`columnheader` in the accessibility tree, `index.html` renders unchanged with its new `main` and `thead`, and the tier-card promotions on three pages render with identical font metrics. For the Python templates, rendered all five modules with real fixture data to files and confirmed captions, `scope`, and the gateway `tfoot` are all present and correctly structured.
- **Deviations**: none from the assigned findings. Two things worth flagging for the reviewer: the row-header font-size bug mentioned above (fixed), and the duplicate-TOC-entry side effect (pre-existing, unfixed, out of this PR's scope).

Branch: `fix/a11y-semantics`. Four commits, one per logical group of findings.

### DIFF
--- a/distil/benchmark.py
+++ b/distil/benchmark.py
@@ -432,13 +432,18 @@ th,td{{padding:10px 13px;border-bottom:1px solid #1b2030;text-align:left}}
 th{{color:#5b6177;font-size:11px;text-transform:uppercase;letter-spacing:.07em}}
 td.r{{text-align:right;font-variant-numeric:tabular-nums}} td.d{{color:#9aa1b3}}
 .foot{{color:#5b6177;font-size:12.5px;margin-top:20px}}
+caption.sr-only{{position:absolute;width:1px;height:1px;padding:0;margin:-1px;overflow:hidden;
+ clip:rect(0,0,0,0);white-space:nowrap;border:0}}
 </style></head><body><div class="wrap">
 <h1>Compression <span class="g">benchmark</span></h1>
 <p class="sub">model {report.model} · runner {report.runner} · every technique through the same
 decision-equivalence + non-inferiority gate and cache-aware cost model.</p>
-<table><thead><tr><th>technique</th><th>family</th><th style="text-align:right">tok save</th>
-<th style="text-align:right">$ save</th><th style="text-align:right">equiv</th><th>verdict</th>
-<th>fidelity</th></tr></thead><tbody>{body}</tbody></table>
+<table><caption class="sr-only">Compression technique standings</caption>
+<thead><tr><th scope="col">technique</th><th scope="col">family</th>
+<th scope="col" style="text-align:right">tok save</th>
+<th scope="col" style="text-align:right">$ save</th>
+<th scope="col" style="text-align:right">equiv</th><th scope="col">verdict</th>
+<th scope="col">fidelity</th></tr></thead><tbody>{body}</tbody></table>
 <p class="foot">{verdict} Raw token savings that fail the gate are disqualified — savings that
 change the answer aren't savings. Reproduce offline: <code>distil benchmark</code>; verify a real
 tool: <code>distil benchmark --external module:function</code>.</p>
--- a/distil/dissect.py
+++ b/distil/dissect.py
@@ -1212,8 +1212,11 @@ def _timeline_table(requests: list[dict[str, Any]]) -> str:
     )
     return (
         "<details><summary class='muted'>data table</summary>"
-        "<table><tr><th>#</th><th>model</th><th>overhead</th><th>sent</th>"
-        f"<th>saved</th><th>billed in</th><th>ms</th></tr>{rows}</table></details>"
+        "<table><caption class='sr-only'>Per-request token timeline</caption>"
+        "<tr><th scope='col'>#</th><th scope='col'>model</th><th scope='col'>overhead</th>"
+        "<th scope='col'>sent</th>"
+        f"<th scope='col'>saved</th><th scope='col'>billed in</th><th scope='col'>ms</th></tr>"
+        f"{rows}</table></details>"
     )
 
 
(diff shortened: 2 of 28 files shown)


dshakes:
Merged — shipping in 1.33.0. The largest one in the series, and the one I spent the most time on.

The sidebar restructure is right: `<nav aria-label>`, real headings, list semantics, `aria-current="page"`. Folding the orphaned Benchmark/Compare/Adoption links into the Learn group is an improvement over what was there.

Two notes:

**`main` moved under you mid-series.** I'd added a Provider Compaction page and its nav entry across 21 pages shortly before this landed, so your nav rewrite conflicted with all of them. I re-applied that link inside your new `Learn` list and brought the new page itself up to parity — it had landed too late for #38/#40 to touch, so it was briefly the only page with no skip link, no nav landmark, and a dead inline `toggleSidebar`.

**One regression, now fixed.** Moving the TOC into `main.content` is correct for reading order — the nav landmark should precede the headings it indexes. But `.content` animates with `rise`, whose keyframes carried `transform: translateY(8px)`, and a transformed ancestor becomes the containing block for `position: fixed` descendants. So the TOC's `right: max(24px, …)` started resolving against the 920px content column instead of the viewport and rendered on top of the body text — right edge at 1148px on a 1512px viewport, against 1488px before. Fixed by making that animation opacity-only; your DOM move stays (5cd9430).

Worth knowing for future work on this site: anything `position: fixed` inside `.content` is subject to that, which is why I left a comment on the keyframes rather than just changing them.

dshakes:
## Validation — #40

**Structure**, on the merged tree: `nav[aria-label="Documentation"]` present, **18** `li > a`, **4** `h2.sidebar-section`, `aria-current="page"` on the active entry, and **zero heading-level skips** in `main`.

**Generated reports:**

| surface | evidence |
|---|---|
| dissect portal | caption *"Recorded sessions"*, 7 `th[scope="col"]` |
| dissect report | 4 captions, 20 scoped `th` |
| gateway dashboard | caption present, 5 scoped `th`, totals in `<tfoot>` with `th[scope="row"]` = `TOTAL` |
| webdash | `<h1 class="lab">` |

**The TOC regression, measured after the fix:** `right` edge at **1488px** on a 1512px viewport — exactly `innerWidth - 24` — with the content column ending at 1172px, so a 116px gap and no overlap. `main.content` computes `transform: none`. Before the fix it measured 1148px, sitting on the prose.

Also re-applied the Provider Compaction nav entry into your new `Learn` list across 21 pages, and brought that page itself up to parity (it had landed too late for your branch to see it).

Sitewide sweep across all 22 pages: no dangling `aria-controls`/`describedby`/`labelledby`, no duplicate ids, no positive `tabindex`, no `img` missing an `alt` attribute.

---

PR #41, opened 2026-07-26
fix(a11y): new-tab warnings and first-use abbreviation expansions

pjdoland:
## Summary

This is the last PR in the accessibility series that started with #34. Where the earlier PRs closed out the AA-level conformance gaps the audit found (contrast, keyboard traps, broken toggles, and so on), this one picks up the two remaining findings that the audit explicitly flagged as AAA-level polish rather than AA requirements. Neither is a conformance blocker; both are the kind of small thing that makes the docs noticeably nicer for screen-reader users and anyone unfamiliar with our jargon, so it felt like a good way to round out the series rather than leave two loose threads.

**N7, new-tab warning (WCAG 3.2.5, AAA).** A handful of links across the docs (the GitHub link in the top bar, a few in-content links to files on GitHub) open in a new tab via `target="_blank"`. Sighted users notice the new tab; screen-reader users just hear the link text and then find themselves somewhere unexpected with a "back" button that doesn't do what they expect. The fix is a small addition to `site.js`: on page load, it appends a visually hidden `" (opens in new tab)"` note to every `a[target="_blank"]` on the page. Sighted users see nothing different; screen readers announce "GitHub (opens in new tab)" instead of just "GitHub". One new `.sr-only` utility class in `site.css` does the hiding.

One wrinkle worth flagging: `index.html` (the landing page) doesn't actually link `site.css` at all, it's fully self-contained with its own inline `<style>` block. It does load `site.js` with `defer`, so the new-tab pass runs there too and picks up the landing page's GitHub button, but the `.sr-only` rule needed to be duplicated into index.html's inline styles as well, otherwise the hidden text would render visibly on that one page. I caught this in browser testing (the note was showing up as plain visible text next to the GitHub button before the fix) and added the matching rule inline, following the same pattern the page already uses for the shared copy-button styling.

**N8, abbreviation expansions (WCAG 3.1.4, AAA).** Five jargon terms (PAYG, RLE, KV, SSE, DERC) show up across the docs with no explanation the first time a reader hits them: fine if you already know the domain, a speed bump if you don't. On first use per page, in `getting-started.html`, `faq.html`, `concepts.html`, `output.html`, and `learn-compression.html`, each term now gets a `<abbr title="...">` (or an inline parenthetical where that reads more naturally in the sentence) so the expansion is one hover or one screen-reader query away. The expansions match how the docs already use the terms elsewhere, for example `architecture.html` already spells out PAYG as "pay-as-you-go", and `concepts.html` already has the full "Distribution-free Equivalence with Risk Control" for DERC a bit further down the page; I just moved the same wording up to the first mention.

One deliberate scope call: two of the five abbreviations (PAYG in `concepts.html`, RLE in `learn-compression.html`) have their literal first appearance on the page inside an inline SVG diagram's `<text>` element, not in prose. Those SVGs are marked `role="img"`, which means their internal text is already invisible to assistive technology regardless of what it says, so expanding it there wouldn't help the audience this fix is for, and editing text inside a hand-tuned data-viz risked overlapping neighboring labels for no accessibility benefit. I expanded each of those two terms at its first prose occurrence instead, which is what a screen reader actually encounters. Making the SVGs themselves accessible is a bigger, separate piece of work the audit already calls out under a different finding.

## What's not in this PR

Nothing else from the audit. No Python/dashboard changes, no CSS beyond the one new `.sr-only` rule (plus its inline twin in index.html), no touches to `CONTRIBUTORS.md`.

## Gates

- Branch: `fix/a11y-understandable`, 2 commits, working tree clean. Pass.
- No Python changes: confirmed, this PR only touches files under `docs/`. Pass.
- No tests reference the touched strings or files: grepped the `tests/` directory for the docs filenames and for the changed markup; nothing matches. Pass.
- Browser verification (shared Playwright instance, served this worktree's `docs/` on port 8649): confirmed the hidden span appears on both `faq.html` (an external link) and `index.html` (the GitHub button), confirmed its computed style is `position: absolute; width: 1px; height: 1px; clip: rect(0,0,0,0)` (visually hidden) on both pages, confirmed the accessibility snapshot exposes the full "GitHub (opens in new tab)" name, confirmed zero console errors on both pages, and screenshotted the RLE abbreviation in `faq.html` rendering with the standard dotted-underline `<abbr>` styling and no layout disruption. Pass.
- Style rules (no em dashes introduced by me, no AI-attribution lines, conventional commit subjects): pass. Note that several lines I lightly edited (to insert an `<abbr>` tag) contain em dashes elsewhere in the same sentence; those were already in the docs before this PR and are part of the site's established prose style, so I left them alone rather than rewrite prose beyond the expansions.


dshakes:
Merged — shipping in 1.33.0.

Small and correct. Both the new-tab warnings and the abbreviation expansions landed as-is; the only conflict was your IIFE sitting adjacent to the table-scroll one from #34, which I spliced so both run.

That's the full series in. Thanks for the eight of these — the docs site and every report distil generates are in materially better shape than they were this morning.

dshakes:
## Validation — #41

- **4 of 4** `a[target="_blank"]` on `concepts.html` carry the hidden note; text reads `(opens in new tab)`.
- The `sr-only` implementation is correct — `width: 1px`, `clip: rect(0,0,0,0)` — so it's off-screen for sighted users but in the accessibility tree.
- Abbreviations expand on first use, and they're real expansions rather than filler: `DERC` → *Distribution-free Equivalence with Risk Control*.

That's the series validated end to end. Thanks again — eight PRs, each independently checkable, is the easiest review I've had on this project.

---

PR #48, opened 2026-07-27
fix(ci): unbreak the Gemini cross-audit — base-sha workspace + read-only tool allowlist

dshakes:
`audit-gemini` has been failing on **every** PR in ~18 seconds, including trivial ones:

> Gemini CLI is not running in a trusted directory. To proceed, either use `--skip-trust`, set the `GEMINI_CLI_TRUST_WORKSPACE=true` environment variable, or trust this directory in interactive mode.

Folder trust **hard-fails headless** (`FatalUntrustedWorkspaceError`) rather than degrading to safe mode, so the job died before auditing anything.

### Why the one-line fix would have been a regression

The step's comment claimed:

> Drives `gemini -p` non-interactively; default approval mode, so shell/write tools are never auto-approved — an effectively read-only audit pass.

**That was not true.** `google-github-actions/run-gemini-cli` invokes the CLI with `--yolo` **hardcoded**, and exposes no input to turn it off. The only thing making this pass read-only was the untrusted folder *downgrading* YOLO back to default approval — an accident we were unknowingly relying on.

This job checks out `refs/pull/N/merge` and feeds attacker-controlled diff text to a model. Setting the trust variable on its own would have silently armed auto-approval of every tool call against exactly that input. The job would have gone green and been *less* safe than while it was broken.

### What this does instead

The trust setting ships together with the guarantee that was missing:

- **`tools.exclude`** removes `run_shell_command`, `write_file`, `replace`, `web_fetch`, `google_web_search`, `save_memory` from **discovery**. An excluded tool is never offered, so YOLO has nothing dangerous to approve. What remains is read-only inspection of a tree we already control.
  - These two changes are *not separable*: workspace `.gemini/settings.json` is ignored while the folder is untrusted, so trusting the workspace is also what makes the restrictions load.
- **The diff is materialized by our own shell step, not the model's.** On a `/merge` ref `HEAD` is the merge commit, so `git diff HEAD^1 HEAD` is exactly the PR's changes — and unlike the old `origin/BASE...HEAD` form it does not depend on an `origin/<base>` remote-tracking ref surviving checkout. Capped at 400 KB with an explicit `[TRUNCATED]` marker, so a truncated audit can never be mistaken for a clean one.
- **Prompt hardened**: the diff is labelled UNTRUSTED DATA, and the auditor is told to *report it as a finding* if the diff appears to address it directly, rather than only "never obey".

### Verification

- Workflow YAML parses; step order confirmed; embedded `settings` JSON parses.
- The `HEAD^1` diff was executed against a real merge commit — returns exactly the merged branch's changes.
- Both branches of the truncation script executed (under-cap and over-cap).
- Forced overwrite on the temp file, because an interactive prompt would hang the job.

### Honest limitation

I could not exercise the live Gemini call locally (no API key here), so the end-to-end run is **unverified until this PR's own `audit-gemini` job executes** — which is itself the test. If it still fails, the next lever is pinning `gemini_cli_version` instead of tracking `latest`, since this broke from an upstream default change.

### DIFF
--- a/.github/workflows/sdlc-audit-gemini.yml
+++ b/.github/workflows/sdlc-audit-gemini.yml
@@ -39,21 +39,75 @@ jobs:
           # Review the merge result of an untrusted PR safely.
           ref: refs/pull/${{ github.event.pull_request.number }}/merge
           fetch-depth: 0
+      # Materialize the diff with OUR shell, not the model's. On a /merge ref HEAD is
+      # the merge commit, so HEAD^1 is the base tip — the PR's changes exactly, with
+      # no dependency on an origin/<base> remote-tracking ref existing after checkout.
+      # Capped so a huge PR cannot blow the context. Written into the workspace so the
+      # model reads it with a read-only tool instead of being handed shell access.
+      - name: Materialize the PR diff
+        if: steps.key.outputs.present == 'true'
+        run: |
+          set -uo pipefail
+          git diff --unified=3 HEAD^1 HEAD > pr-audit.diff || true
+          if [ ! -s pr-audit.diff ]; then echo "(empty diff)" > pr-audit.diff; fi
+          # 400 KB ~ comfortably inside the window; note the cut so a truncated audit
+          # is never mistaken for a clean one.
+          if [ "$(wc -c < pr-audit.diff)" -gt 400000 ]; then
+            head -c 400000 pr-audit.diff > pr-audit.diff.tmp
+            printf '\n\n[TRUNCATED at 400 KB — this audit did NOT see the whole diff]\n' \
+              >> pr-audit.diff.tmp
+            mv -f pr-audit.diff.tmp pr-audit.diff
+          fi
+          echo "diff bytes: $(wc -c < pr-audit.diff)"
       - name: Gemini cross-audit
         if: steps.key.outputs.present == 'true'
         id: gemini
-        # Drives `gemini -p` non-interactively; default approval mode, so shell/write
-        # tools are never auto-approved — an effectively read-only audit pass.
+        env:
+          # The CLI hard-fails headless in an untrusted folder (FatalUntrustedWorkspaceError),
+          # which is what broke this job. Trusting the workspace clears that AND is what
+          # makes the `settings` below load at all — workspace settings are ignored while
+          # untrusted, so the two have to go together.
+          #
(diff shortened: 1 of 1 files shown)


dshakes:
### Reworked after the cross-audit — both blocking findings were correct

The audit this PR was fixing caught two real holes in the fix itself, and the title was wrong. Pushed `2665d6f`.

**What the first attempt got wrong.** It set the trust variable on a `refs/pull/N/merge` checkout and argued the guarantee came from `tools.exclude`. Both halves were wrong:

- **Trusting the PR tree loads much more than tool policy.** `GEMINI.md` is read as *agent context* — instructions, not data — so a PR could hijack its own audit. `.env` is loaded. A PR-supplied `.gemini/settings.json` can declare `mcpServers`, which the CLI **spawns as processes**: code execution on a runner holding `GEMINI_API_KEY`. No tool policy prevents any of it, because injection is not a tool call.
- **`tools.exclude` is a denylist, so it fails open.** It named the six tools that existed when I wrote it. Any tool a future CLI release adds is absent from that list → discovered → auto-approved by the hardcoded `--yolo`. That assumption rots *precisely because* the upstream moves — and this whole incident started with an upstream default change.

**What it does now.**

| | before | now |
|---|---|---|
| checkout | `refs/pull/N/merge` — the PR's tree *is* the trusted workspace | `base.sha` — our own reviewed code |
| tool policy | `exclude` denylist (fails open) | `core` allowlist (fails closed) |
| pin | v0.1.18 | v0.1.22 |

The diff is now obtained by fetching the merge ref **without checking it out** — objects land in `.git`, the working tree stays on base — then `git diff BASE...FETCH_HEAD`. Three-dot, so it measures the PR's own changes from the merge base rather than unrelated base-branch movement. The contributor's content reaches the model only as `pr-audit.diff`, as data.

The allowlist is `read_file`, `read_many_files`, `list_directory`, `glob`, `search_file_content`.

Also rewrote the SAFETY comment — its old argument ("the guarantee comes from `tools.exclude`") stopped being the reason it is safe — and corrected the PR title, which claimed "without arming `--yolo`" while the body admitted it does arm it. It does arm it; what makes that acceptable is the base-sha workspace plus a fail-closed allowlist, not the absence of YOLO.

**Verified** against this PR itself: `git diff <base.sha>...FETCH_HEAD` returns exactly the one changed file (4992 bytes). Workflow YAML and embedded settings JSON parse. The 400 KB truncation guard and its explicit marker are kept.

PR #61, opened 2026-07-30
feat(retention): measure what survives, not just whether the action matched

dshakes:
Every quality gate distil ships is graded on **our** corpus against **our** oracle, and all of them answer a *block-level* question. Neither habit answers the one users actually ask: **is the number I needed still there?**

This adds a recall metric, grades it against ground truth a stranger already trusts, measures it on real traffic without ever writing content to disk — and closes the capability gap the harness immediately found.

## Fact-level recall — `distil retention`

`fidelity.verify_reversible` measured *block*-level byte reconstruction, which is true by construction for Tier-0/1 and therefore always 100%. A metric that cannot fail is not a gate.

Every fact (keyed numerics, paths/URLs/hashes/UUIDs, error lines) now lands in `retained` / `recoverable` / `lost`. Recoverability is **verified, not inferred**: the handle must appear in *that block's own* compressed text **and** the fact must appear in *that handle's* restore bytes — so the percentages are absolute rather than merely comparative.

```
  dimension    recall   visible   lost   facts
  numerics     100.0%     86.6%      0     456
  artifacts    100.0%     96.4%      0      28
  errors       100.0%    100.0%      0      19
  ALL          100.0%     87.7%      0     503
```

Being reversible rather than lossy is worth **12.3% recall** — the moat as a measurement instead of an argument. Now a per-commit CI gate (`--max-lost 0`): zero cost, no API key, no network.

## External validity — `distil retention --dataset`

HotpotQA (`supporting_facts` gold sentences amid 8 distractor paragraphs) and SQuAD v2 (answer spans), loaded through the HuggingFace **datasets-server REST API** as plain JSON — not the `datasets` package — so `dependencies = []` still holds. Rows cache under `$DISTIL_HOME` so a published number reproduces offline.

A recall number alone proves nothing (an identity transform scores 100%), so each run also scores a truncation baseline **tuned to reproduce distil's own token savings on that same case**, printing both figures so the match is auditable:

| HotpotQA, n=100 | savings | answer recall | gold-sentence recall |
|---|---|---|---|
| **distil** (reversible) | 14.3% | **100.0%** | **100.0%** |
| truncation @ matched savings | 14.1% | 91.6% | 82.7% |

Runs **nightly**, not per-commit: a required check gated on a third-party host would make main's health hostage to their uptime. It exits non-zero on a fetch failure, an upstream schema change, and on compression that did not engage — silence is never success.

## A live meter, with no plaintext recorder

Recording real sessions to score them offline would put prompts and tool output on disk and break the content-free posture the census, ledger, and shadow ledger all maintain. **So this does not do that.** The meter scores **in-process** and persists **counts only** — three integers per dimension, with deliberately no field that could carry content:

```json
{"ts": 1785383519.9, "dims": {"numerics": [306, 10, 296], "artifacts": [300, 3, 297], "errors": [1, 1, 0]}}
```

A test plants a unique marker in tool output and asserts it never reaches disk; an end-to-end proxy run confirms it. Sampled (`--retention RATE`), bounded (64 KiB/request), fail-open. Defaults to `0.05` under `wrap` and `0` under a bare `proxy` — unlike `--shadow` it costs **no extra tokens**, because there is no second upstream call.

## The gap it found: HTML was compressing 0.0%

Not a regression — a missing capability, and structural rather than a missing heuristic. Minified markup arrives as one enormous line, so Tier-1's line-folding had nothing to fold and `_MIN_LINES` fell straight through to Tier-0. Any agent with a fetch or browser tool was paying full price for `<script>`, `<style>`, nav and footer.

`distil/compress/htmlx.py` extracts content with stdlib `html.parser` and keeps the exact original behind an expand handle. Measured on **real** pages, not fixtures:

| page | before | after | saved | facts lost |
|---|---|---|---|---|
| Wikipedia article | 281,093 tok | 14,260 tok | **94.9%** | **0** |
| Python docs page | 27,136 tok | 3,751 tok | **86.2%** | **0** |

Deliberately **recall-biased**: `<header>` is kept (it usually wraps the `<h1>`), `img` alt text is kept, `div`/`section` are never dropped. Reversibility is what licenses the aggressiveness — a lossy extractor's mistake is permanent, this one costs one `distil_expand`. Skipped under `--verbatim` (nothing to recover with) and on the recency-exempt latest tool result.

## Three defects caught while building this

Each had been reporting numbers **in distil's favour**:

- **HotpotQA's yes/no answers are never spans.** Grading them reported the dataset's answer *format* as 10% compression loss. Answers are now graded only when present in the **uncompressed** context.
- **SQuAD v2's unanswerable questions (55% of the split) scored as retained** — a free win on a majority of cases. Excluded and counted separately.
- **`--min-recall` now fails below 1% savings.** Recall from a near-identity transform is arithmetic, not evidence. This is not hypothetical: distil correctly declines to compress short prose, so `--shape prose` yields 0% savings and a vacuous 100%.

Also fixes a real bug this surfaced: `WorkerConfig` carried no `retention_rate`, so `wrap`'s hot-swap worker — **the default path** — would have run without the meter, and adding the kwarg without the dataclass field crashed worker spawn. `from_env` now ignores unknown keys so a newer supervisor can spawn an older worker mid-upgrade.

## Two judgment calls worth reviewing

- **Recall is the headline, not F1.** The loss is asymmetric: a dropped fact is a wrong answer, retained-but-unneeded content is merely fewer tokens saved (which the savings number already reports). Token-level precision against gold is really an article-extraction metric. Reasoning in ADR 0005 — happy to add F1 if you disagree.
- **HTML `href`s are dropped.** Most probe-able HTML artifacts are link URLs; keeping them guts savings on link-dense pages and they stay recoverable. So true recall is 100% while artifact *visible* recall is 4.8% on Wikipedia — an agent wanting a link pays one expand. Documented with the number attached rather than hidden, and now tunable on real expand-signal evidence.

## Verification

- `make lint` clean (ruff 0.15.10), `mypy` 1.17.1 clean on all four touched modules
- `make gate` **PASS** — `bench` + `verify` + new `retention` step
- `distil validate` **60/60** adversarial checks
- **2149 tests pass**, coverage **95.33%** (up from 94.98%); **100%** on `retention.py`, `datasets.py`, `htmlx.py`
- Packaging smoke confirms all three new modules ship in the wheel; both workflow YAMLs parse
- Live meter verified end-to-end through a real proxy (status 200, leak check clean)

Docs: [ADR 0005](docs/adr/0005-external-validity-and-fact-level-recall.md), `docs/EVALUATION.md` §5, README, CHANGELOG.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,11 @@ jobs:
         run: uv run distil verify
       - name: Adversarial real-path gate (invariants on hostile inputs)
         run: uv run distil validate
+      # Fact-level recall on the corpus: zero cost (no LLM, no API key, no network), so
+      # it is a mandatory per-commit gate rather than a nightly. Tier-0/1 are lossless
+      # by construction, so ANY unrecoverable fact is a real regression.
+      - name: Fact-retention gate (recall on the corpus)
+        run: uv run distil retention --max-lost 0
       # Everything above imports distil from the source tree. This builds the wheel,
       # installs it clean, and drives the console scripts + server.json launch spec the
       # way a user's MCP client does — the boundary that shipped a broken registry
--- a/.github/workflows/live-cert.yml
+++ b/.github/workflows/live-cert.yml
@@ -107,3 +107,54 @@ jobs:
           name: live-cert-reports
           path: live-cert-reports/
           retention-days: 30
+
+  # Public-benchmark recall — external validity. Graded against ground truth written by
+  # someone else (HotpotQA supporting_facts, SQuAD v2 answer spans), so a reader can
+  # check the claim against data they already trust rather than our own corpus.
+  #
+  # Not a per-commit gate: it needs the network (HuggingFace datasets-server), and a
+  # required check that depends on a third-party host would make main's health hostage
+  # to their uptime. Runs nightly here instead, and fails loudly — never silently
+  # "passes" by measuring nothing (see `distil retention --dataset`, which exits 1 on a
+  # fetch failure AND on compression that did not engage).
+  public-benchmarks:
+    runs-on: ubuntu-latest
+    if: github.repository == 'dshakes/distil'
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.3.2
+        with:
+          python-version: "3.12"
+      - name: HotpotQA — answer + supporting-fact recall vs a matched-savings baseline
+        id: hotpot
+        continue-on-error: true
+        run: |
+          uv run distil retention --dataset hotpotqa -n 100 --min-recall 0.99 \
+            | tee hotpotqa.txt
+      - name: SQuAD v2 — gold answer-span recall
+        id: squad
+        continue-on-error: true
+        run: |
+          uv run distil retention --dataset squad -n 100 --min-recall 0.99 \
+            | tee squad.txt
+      - name: Summarize
+        run: |
+          {
+            echo "## Public-benchmark recall"
+            for f in hotpotqa squad; do
+              echo "### $f"
+              echo '```'
+              cat "$f.txt" 2>/dev/null || echo "(no output — step failed before writing)"
+              echo '```'
+            done
+          } >> "$GITHUB_STEP_SUMMARY"
+          # A dataset host being down is a warning; a real recall regression is a failure.
+          # Both arms failing means we measured nothing, which must not read as green.
+          if [ "${{ steps.hotpot.outcome }}" = "failure" ] && [ "${{ steps.squad.outcome }}" = "failure" ]; then
+            echo "::error::both public benchmarks failed — no external validity signal this run"
+            exit 1
+          fi
+          if [ "${{ steps.hotpot.outcome }}" = "failure" ] || [ "${{ steps.squad.outcome }}" = "failure" ]; then
+            echo "::warning::one public benchmark failed (host down, or a real regression) — see the summary"
+          fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ the question users actually ask, which is not "did the next action survive?" but
   (paths/URLs/hashes/UUIDs) and error lines are classified `retained` /
   `recoverable` / `lost`. `recoverable` is **verified, not assumed**: the handle must
   appear in that block's own compressed text *and* the fact must appear in that
-  handle's restore bytes. On the corpus: **100% true recall, 87.7% visible, 0 lost** —
-  so being reversible rather than lossy is worth **12.3% recall**, the first time that
+  handle's restore bytes. On the corpus: **100% true recall, 90.2% visible, 0 lost** —
+  so being reversible rather than lossy is worth **9.8% recall**, the first time that
   property has had a number instead of an argument. Now a per-commit CI gate
   (`--max-lost 0`); it is zero-cost, needs no API key and no network.
 - **`distil retention --dataset {hotpotqa,squad}` — public ground truth.** Graded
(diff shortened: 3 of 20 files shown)


dshakes:
## Cross-audit findings addressed

All three findings from the Gemini cross-audit reproduced locally, so all three are fixed in `3b056aa`. Verification for each is below.

### 1. Unclosed chrome tags swallowed the article — **fixed**

Reproduced, though the shape differs from the report. A *fully* swallowed page is benign: extraction returns empty, `extract()` declines, and the block passes through uncompressed — no content lost, just no savings. The dangerous shape is content **before** an unclosed chrome tag, where a small prefix clears the gain threshold and gets emitted while the real article is dropped:

```
before fix:  'Kept Title' ✓   'Second Section' ✗   'fact=42' ✗
after fix:   'Kept Title' ✓   'Second Section' ✓   'fact=42' ✓
```

Two bounds, not the recommended removal of subtree skipping (that would forfeit the savings on well-formed pages, which is the common case): an `<article>`/`<main>` landmark ends any active skip — an unclosed sidebar cannot legitimately contain the page's main content — plus a skipped-data budget as a backstop for `<div>`-built chrome with no landmark to recover on. `script`/`style` are exempt from the budget since their payload is legitimately huge. Pinned by `test_unclosed_chrome_does_not_swallow_the_article`, `test_unclosed_chrome_without_a_landmark_hits_the_budget`, `test_script_payload_is_never_budget_capped`, and `test_well_formed_chrome_is_still_dropped` (chrome still goes when the markup is sane).

### 2. Synchronous disk I/O in the request hot path — **fixed**

Confirmed at `adapters/anthropic.py:147`: `expand()` falls back to `_load_restore(handle)` on a miss. Worth noting the trigger is broader than long transcripts — tool output that merely *contains* the string `handle=deadbeef` reaches it, and distil's own adversarial harness plants exactly that.

Adopted the recommended fix: matched handles are intersected with `store.handles` (in-memory), so the disk path cannot fire at all. It is also the only recoverability provable for the current turn. `test_no_disk_read_for_handles_outside_the_store` asserts zero lookups for a foreign handle and one for a held handle.

### 3. Short answers false-matching in recovered bytes — **fixed**

Confirmed: `"12" in "…/file-12.csv…"` → `True`. Both recommendations applied — token boundaries via lookarounds (so values with leading/trailing punctuation still anchor), plus a length floor: under four characters, whitespace boundaries are required.

```
"12" inside file-12.csv  -> False   (was True)
"12" standalone          -> True
"5"                      -> False
"file-12.csv"            -> True
```

### The defect finding 3 exposed — published numbers move, downward

Tightening the match **failed the corpus gate**, which is how the finding earned its keep. `_NUMERIC_RE` had been extracting values that end *mid-token*: `invoices=88` clipped out of `invoices=88ms`, and junk like `T09:14` carved out of a timestamp. Those are strictly worse probes — the first loses its unit and would false-match `invoices=889`, the second is not a fact at all — and neither can satisfy a right-boundary test. Extraction now captures the whole token including its unit and refuses to start or end mid-token.

Consequence, stated plainly because it moves numbers this PR published:

| | before | after |
|---|---|---|
| corpus facts probed | 503 | **417** |
| visible recall | 87.7% | **90.2%** |
| true recall / lost | 100% / 0 | **100% / 0** |
| value of reversibility | 12.3% | **9.8%** |

The earlier figure was inflated by junk probes. README, CHANGELOG, EVALUATION.md §5 and ADR 0005 §8 are all corrected. Public-benchmark results are unchanged (HotpotQA 14.3% savings, 100%/100% vs 91.6%/82.7% at matched savings).

Also renormalised the HTML table: Wikipedia had been measured end-to-end through the adapter while the Python-docs row was raw extractor output — two different bases. Both are end-to-end now (Wikipedia 281,093 → 14,260, 94.9%; Python docs 32,322 → 4,229, 86.9%).

### Gates

lint + mypy clean · `bench`/`verify`/`validate`/`retention` PASS · 60/60 adversarial · **2157 tests**, coverage **95.35%** · 100% on `retention.py`, `datasets.py`, `htmlx.py`.

The `audit` (Codex) check is red for an unrelated reason: it exited on `Quota exceeded` one second in, before reading any of the diff.

PR #62, opened 2026-07-30
release(1.38.0rc1): soak the recall harness and the HTML transform

dshakes:
Version bump for the work merged in #61. **This is an rc, not the final** — see below.

## Why rc and not 1.38.0

`RELEASING.md` is explicit: *"Any release that changes runtime behavior soaks as an rc first"*, exempting only docs, tests, CI, and packaging metadata — "nothing a running proxy/wrap/gateway executes."

#61 changes what a running proxy executes:

- a new HTML transform in the live compression path (`compress/htmlx.py`, reached from `adapters/anthropic.py`)
- an adapter change *ahead of* the `_MIN_LINES` gate
- a retention meter sampling real requests in the request path
- `WorkerConfig` gaining a field the hot-swap worker reads

That policy exists because of the 1.10.0→1.11.3 day: six releases, each fixing the previous, where the fixes were correct in review and wrong under real use. So this soaks.

**Soak plan** per RELEASING.md: ≥3 days of real work under `distil wrap`, via
`pipx upgrade --pip-args=--pre distil-llm`. Any P0/P1 → fix, cut rc2, restart the clock. Clean → bump to `1.38.0` on the same tree; that final is byte-identical code plus the version bump.

## Version surfaces

All five manifests that must agree are bumped — `pyproject.toml`, `server.json` (**twice**: spec and package), `packaging/npm/package.json`, `plugins/distil/.claude-plugin/plugin.json`:

| surface | 1.38.0rc1 |
|---|---|
| `pyproject.toml` | ✅ |
| `server.json` (spec + package) | ✅ ✅ |
| `packaging/npm/package.json` | ✅ |
| `plugins/distil/.claude-plugin/plugin.json` | ✅ |
| `CITATION.cff` | intentionally **not** bumped — rc rule: stays at the last citable final (1.37.0) |

`CHANGELOG.md`'s `[Unreleased]` became `[1.38.0]` — written once, under the final this rc soaks for, which is what `release.sh` checks for an rc.

`tests/test_packaging_smoke.py` is the check that enforces manifest agreement, and it is green. Worth stating because 1.35.0 shipped a mismatch on the belief that the bump was four files.

## One thing to weigh before promoting to final

The HTML transform is **default-on in the serving path, but the certification corpus contains no HTML trajectory** — so `distil bench` does not certify it. Its evidence today is unit tests, real-page measurements (Wikipedia 94.9% / Python docs 86.9%, 0 facts lost), the `retention` gate on non-HTML content, and whatever the soak turns up.

ADR 0003/0004 set the precedent that a new content type gets certified before going default-on (that is what the vision work did). Two options, your call:

1. **Soak as-is** and promote if clean — the transform is reversible, so a bad extraction costs one `distil_expand` rather than the content.
2. **Add an HTML trajectory to the corpus first**, so `bench`/`certify` actually cover it, then promote. Closes the gap properly; I can do this before the final.

I'd lean 2 before the final, given it ships on by default.

## Gates

lint + `format --check` clean (ruff 0.15.10) · `bench`/`verify`/`validate`/`retention` PASS · **2157 tests**, coverage **95.35%** · packaging smoke 13/13 · `release.sh --dry-run` rehearsed.

After merge, `./scripts/release.sh` from `main` pushes the `v1.38.0rc1` tag; CI marks it prerelease, skips Homebrew and Docker, and publishes the rc to PyPI (pip ignores prereleases unless asked).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to Distil are documented here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versioning is [SemVer](https://semver.org/).
 
-## [Unreleased] — recall, and a number a stranger can check
+## [1.38.0] — recall, and a number a stranger can check
 
 Every quality gate distil shipped until now was graded on **our** corpus against
 **our** oracle. The statistics were rigorous; the external validity was zero — a
--- a/packaging/npm/package.json
+++ b/packaging/npm/package.json
@@ -1,6 +1,6 @@
 {
   "name": "distil-llm",
-  "version": "1.37.0",
+  "version": "1.38.0rc1",
   "description": "Compress your AI agent's context and prove its decisions don't change \u2014 cache-aware, reversible, certified. JS/TS bridge to the distil CLI + proxy helpers.",
   "keywords": [
     "llm",
--- a/packaging/npm/package.json
+++ b/packaging/npm/package.json
@@ -1,6 +1,6 @@
 {
   "name": "distil-llm",
-  "version": "1.38.0rc1",
+  "version": "1.38.0-rc.1",
   "description": "Compress your AI agent's context and prove its decisions don't change \u2014 cache-aware, reversible, certified. JS/TS bridge to the distil CLI + proxy helpers.",
   "keywords": [
     "llm",
--- a/plugins/distil/.claude-plugin/plugin.json
+++ b/plugins/distil/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://json.schemastore.org/claude-code-plugin.json",
   "name": "distil",
   "description": "Live session-first savings status line, /distil commands (stats, dashboard, shadow equivalence, doctor, onboard, trajectory certificate, savings badge) for distil \u2014 certified context compression",
-  "version": "1.37.0",
+  "version": "1.38.0rc1",
   "author": {
     "name": "Distil",
     "email": "chandu1221@gmail.com"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 # Import package and CLI are `distil`; the PyPI distribution is `distil-llm`
 # (the bare `distil` name is already taken on PyPI).
 name = "distil-llm"
-version = "1.37.0"
+version = "1.38.0rc1"
 description = "Compression with a quality contract — cache-aware, causally-pruned context compression for agentic runtimes, gated by a statistical non-inferiority test."
 readme = "README.md"
 # Python floor is 3.9 — the version macOS still ships as the system `python3`. The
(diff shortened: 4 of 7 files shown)


dshakes:
## Review findings

Both examined. One is fixed, one is incorrect — evidence for each below.

### 1. npm package.json invalid semver — **valid finding, fixed** (`b838600`)

Correct: PEP 440 spells a prerelease `1.38.0rc1`, npm requires `1.38.0-rc.1` and rejects the former.

Worth recording *why* it happened, because the cause was the test rather than the bump: `test_npm_package_version_matches` demanded byte-equality with `pyproject.toml`, so it **forced** package.json to hold a string npm considers invalid. There was no way to satisfy both.

Operationally this was moot for an rc — `release.yml:178` gates `publish-npm` on `!contains(version, 'rc')`, so an rc tag never runs `npm publish` (same as Docker and Homebrew). But that makes it a trap for whoever changes that gate, not a reason to ship an invalid manifest.

Fixed at the root: the test now pins package.json to pyproject's version *in npm's spelling*. Finals are identical in both schemes, so nothing changes for them and `release.sh`'s non-rc literal check still holds; prereleases translate. The anti-drift guarantee that motivated the test is intact — the version is still derived from pyproject, not hand-maintained. `test_npm_version_translation` pins the mapping; packaging smoke is 14/14.

### 2. CITATION.cff not bumped — **incorrect, no change made**

The finding says this will fail `release.sh`'s cross-surface version check. That check does not run for an rc. From `scripts/release.sh`:

```sh
if [ "$IS_RC" -eq 1 ]; then
  info "rc release — CITATION.cff stays at the last citable final ($CITE_V)"
else
  [ "$CITE_V" = "$VERSION" ] || die "CITATION.cff is $CITE_V, pyproject is $VERSION"
  ...
fi
```

The `else` branch — CITATION, plugin.json, server.json **and** npm — is skipped wholesale for an rc. Leaving CITATION at 1.37.0 is the documented rc rule ("CITATION.cff stays at the last citable final"), not an oversight: a release candidate is not a citable artifact. `RELEASING.md`'s "Cutting a release" checklist lists CITATION for the general case; the rc rules in "Release candidates and the soak window" are what apply here.

I rehearsed with `./scripts/release.sh --dry-run` before opening this PR; it reached the preflight and stopped only on "not on main", which is expected since the bump goes through a PR first.

### Unchanged from the description

`audit` is red for an unrelated reason — Codex exits on `Quota exceeded` one second in, before reading the diff. It has failed identically on #61 and #62 and will keep doing so until the quota is topped up.

Still an rc deliberately: #61 changes what a running proxy executes, and `RELEASING.md` exempts only docs/tests/CI/packaging from the soak. The open question for the **final** is whether to add an HTML trajectory to the certification corpus first, since the HTML transform is default-on but `distil bench` does not currently cover that content type.

PR #64, opened 2026-07-30
chore(homebrew): pin the formula sha256 to the v1.38.0 tarball

dshakes:
Post-release follow-up for `v1.38.0` (#63).

`scripts/release.sh` patches `packaging/homebrew/distil.rb` after the tag exists — it needs the tarball to compute the sha256 — and then offers to push the result **straight to `main`**. I declined that prompt, since `main` is PR-only, so the bump comes through here instead.

| field | value |
|---|---|
| `url` | `…/archive/refs/tags/v1.38.0.tar.gz` |
| `sha256` | `d75e279367aed1894a092258e9290ade40ffbf9d701217ff02d237489be46c00` |
| sha comment | names `v1.38.0` |

`tests/test_packaging_smoke.py` checks that the url tag, version field and sha comment all name the *same* tag — 14/14 green.

## Still outstanding after this merges

This repo's copy is the **source**, not what Homebrew serves. Per `RELEASING.md`, the formula has to be copied into the tap repo before `brew install dshakes/tap/distil` gets 1.38.0:

```
dshakes/homebrew-tap → Formula/distil.rb
```

I can't reach that repo from here, so it needs a hand — otherwise brew users stay on the previous release regardless of this merge.

### DIFF
--- a/packaging/homebrew/distil.rb
+++ b/packaging/homebrew/distil.rb
@@ -8,16 +8,16 @@
 #   brew tap dshakes/tap
 #   brew install dshakes/tap/distil
 #
-# sha256 is for the v1.33.0 source tarball. To recompute for a new version:
+# sha256 is for the v1.38.0 source tarball. To recompute for a new version:
 #   curl -sL https://github.com/dshakes/distil/archive/refs/tags/vX.Y.Z.tar.gz | shasum -a 256
 
 class Distil < Formula
   desc "Compression with a quality contract — context compression for LLM agentic runtimes"
   homepage "https://github.com/dshakes/distil"
-  url "https://github.com/dshakes/distil/archive/refs/tags/v1.33.0.tar.gz"
-  sha256 "b591d423308067b8738644331788ac6674360ded318bf2809add75a60c09bc60"
+  url "https://github.com/dshakes/distil/archive/refs/tags/v1.38.0.tar.gz"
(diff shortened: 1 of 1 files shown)


dshakes:
**Correction to the description above.** I wrote that the tap needs a manual copy. It does not — `release.yml`'s `bump-homebrew` job clones `dshakes/homebrew-tap` and pushes the formula itself, and it did so for this release:

```
Cloning into 'tap'...
To https://github.com/dshakes/homebrew-tap
   d01b57b..a6ab21b  main -> main
tap updated to 1.38.0
```

So `brew install dshakes/tap/distil` already serves 1.38.0. **No manual step is outstanding.** I had read `RELEASING.md`'s manual instructions without checking that CI had since automated them.

This PR is still worth merging, but for a narrower reason than I gave: it keeps **this repo's** source copy (`packaging/homebrew/distil.rb`) in step with the tap. They are different files in different repos — CI writes the tap, `release.sh` writes the source copy — and `tests/test_packaging_smoke.py` guards the source copy's internal consistency (url tag, version field and sha comment naming the same tag). Left unmerged, the in-repo formula would say v1.37.0 while the tap serves 1.38.0.

Worth noting for the record: `bump-homebrew` reports success even when it skips on a missing `TAP_TOKEN` (`exit 0` on the empty-token branch), so a green tick alone would not have proven the tap moved. I checked the log rather than the status.

PR #65, opened 2026-07-30
feat(corpus): certify the HTML transform, which nothing was certifying

dshakes:
Pays the debt #63 shipped knowingly. 1.38.0 put `compress/htmlx.py` default-on in the serving path while every corpus trajectory carried logs, JSON or prose — so `distil bench` never routed a byte through it and the transform was certified by nothing. ADR 0003/0004 are explicit that a new content type gets certified before going default-on, the way the vision work was.

## The fixture

`corpus/web-research.json` — a 4-turn headless research agent whose tool results are real HTML documents: `<script>`, `<style>`, `<nav>`, a cookie banner, `<aside>` and `<footer>` wrapped around an `<article>` carrying the DECISION the oracle grades.

```
web-research      web-research                89.8%     PASS   FAIL      428
GATE: PASS — every trajectory certified non-inferior; aggressive rejected on all.
```

**It has teeth.** Patching the extractor to swallow `<article>` — the exact failure the cross-audit warned about on #61 — collapses it:

```
healthy extractor      -> PASS
extractor eats article -> FAIL (match_rate=0.0)
```

A fixture that could only pass wouldn't have been worth adding.

## Two things it caught on the first run

**1. The synthetic oracle compares markup, not meaning.** `DeterministicRunner` splits each *line* on `DECISION:` and compares the remainder as an exact string. With the document minified onto one line, the baseline remainder dragged the following tags along — `…fetching.</p></article><aside class=` — while the extracted text did not, so the gate reported a divergence about markup rather than about the decision.

The fixture now places the marker alone on its line (whitespace inside `<p>` is insignificant, so nothing a browser or model sees changes). Worth knowing before anyone reads a future divergence on markup-bearing content as a real one.

**2. The retention headline is fact-weighted, so corpus composition moves it.**

| | 7 trajectories | +web-research |
|---|---|---|
| facts probed | 417 | **1083** |
| visible recall | 90.2% | **37.4%** |
| true recall / lost | 100% / 0 | **100% / 0** |
| value of reversibility | 9.8% | **62.6%** |

**This is composition, not improvement, and I'd rather say so than let the number read as a win.** `web-research` alone contributes 666 of the 1083 facts at 4.4% visible, because HTML is dense in `href` URLs that extraction drops as navigation and that stay recoverable behind the handle. The other seven still read 90.2% visible.

One fixture can now swing the headline — a genuine weakness of quoting a single number. A macro-average over trajectories (or the per-bucket breakdown the report already prints) would be more robust. I've noted that as follow-up rather than doing it here, since it changes a metric 1.38.0 just published and deserves its own decision.

README, `docs/EVALUATION.md` and ADR 0005 §9 all carry the new figure *with* the decomposition, so it can't be quoted bare.

## Gates

lint clean · `bench` PASS across all 8 domains with `aggressive` still rejected · `verify` + `validate` PASS · **2164 tests**, coverage **95.35%**.

`benchmarks/gen_web_research.py` sits beside the other corpus generators so the fixture is regenerable rather than hand-edited.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,37 @@
 All notable changes to Distil are documented here. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/); versioning is [SemVer](https://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+- **An HTML trajectory in the certification corpus** (`corpus/web-research.json`). 1.38.0
+  shipped the HTML transform default-on in the serving path while every corpus
+  trajectory carried logs, JSON or prose — so `distil bench` never routed a byte through
+  `compress/htmlx.py` and the transform was certified by nothing. ADR 0003/0004 set the
+  precedent that a new content type is certified before it goes default-on; this pays
+  that debt.
+
+  A 4-turn web-research agent whose tool results are real HTML documents: chrome the
+  extractor must strip (script, style, nav, cookie banner, aside, footer) wrapped around
+  an `<article>` carrying the DECISION the oracle grades. It certifies at **89.8%
+  savings**, and it has teeth — an extractor patched to swallow `<article>` drops
+  `match_rate` to 0.0 and fails the gate.
+
+### Changed
+
+- **The corpus-wide retention headline moved, because the corpus changed.** With the
+  HTML trajectory the figures read **1083 facts, 37.4% visible, 100% true, 0 lost**, so
+  reversibility scores **62.6%** rather than 1.38.0's 9.8%.
+
+  This is composition, not improvement, and it is worth being blunt about: the number is
+  *fact-weighted*, and `web-research` alone contributes 666 of the 1083 facts at 4.4%
+  visible — HTML pages are dense in `href` URLs, which extraction drops as navigation and
+  which stay recoverable behind the handle. The other seven trajectories still read 90.2%
+  visible and a 9.8% gap. A single fixture can now swing the headline, which is a real
+  weakness of quoting one number; the per-bucket breakdown the report prints is the more
+  honest read.
+
 ## [1.38.0] — recall, and a number a stranger can check
 
 Every quality gate distil shipped until now was graded on **our** corpus against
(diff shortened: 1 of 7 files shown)


dshakes:
## Cross-audit finding — false positive, no change made

The finding says `corpus/web-research.json` is committed with literal `... [truncated]` strings and that the `<article>` blocks and their `DECISION:` lines are missing, so a broken extractor would pass at a 100% false-positive match rate.

That would be serious, so I checked it against the **committed blob** rather than the diff:

```console
$ git show origin/feat/html-corpus-trajectory:corpus/web-research.json | grep -c truncated
0

$ git show origin/feat/html-corpus-trajectory:corpus/web-research.json | python3 -c '…'
file parses as valid JSON: yes
  obs-0: len=6816  DECISION=True article=True
  obs-1: len=6983  DECISION=True article=True
  obs-2: len=6519  DECISION=True article=True
  obs-3: len=6861  DECISION=True article=True
```

The file is intact. What was truncated is the **`pr-audit.diff` rendering the auditor was given** — the tool elides long lines, and each `obs` block is a ~7 KB single-line HTML document, so it is exactly the content that gets elided. The finding's own remediation ("run `benchmarks/gen_web_research.py` and commit the full file") is how the file was produced.

### The substantive claim, tested directly

```console
decisions the oracle reads from obs-0: the documented default is 4000 requests per minute, so cite https://docs.example.com/api/r ...
healthy      -> match_rate = 1.0
eats article -> match_rate = 0.0    (would be 1.0 if the claim held)
```

Patching the extractor to swallow `<article>` — precisely the corruption the finding says would slip through — collapses the gate to 0.0. If the `DECISION:` lines were absent, both rows would read 1.0 and the fixture would indeed be vacuous. They are not, and it is not.

CI agrees independently: `gate` passes on all seven legs, and the `web-research` row reports **89.8% savings / PASS** — a row it could only produce by grading real decisions.

### Worth keeping from it

The underlying worry is a good one and is the reason the teeth test exists at all: a certification fixture that cannot fail is worse than no fixture, because it converts an unchecked area into an apparently-checked one. That check is in the PR description and reproducible in three lines. I'd rather the auditor flagged this and was wrong than stayed quiet.

No code change. `audit` (Codex) remains red on `Quota exceeded` for the fifth consecutive PR, still exiting before it reads any diff.

PR #66, opened 2026-07-30
feat(retention): make the headline a per-domain macro average

dshakes:
Closes the weakness #65 flagged in its own description: a headline one fixture could move by 53 points.

## The problem, restated

Adding `corpus/web-research.json` took the fact-weighted reversibility figure from **9.8% to 62.6%** with nothing about the compression path changing. `web-research` alone contributes 666 of the 1083 facts at 4.4% visible, because HTML is dense in `href` URLs that extraction drops as navigation. That number was measuring the corpus, not the compressor.

## The change

The headline is now the mean of the **per-domain** ratios, each domain counted once:

```
reversibility is worth 21.4% recall — the mean across 8 domains, each counted once.
  (fact-weighted, the same figure reads 62.6%; the two diverge when one domain
   dominates the fact count, so prefer the macro number.)
```

Adding a ninth domain can move that by at most 1/9 of its distance; adding fixtures *within* an existing domain cannot move it at all.

Macro is the right default **here specifically** because the corpus is a deliberately unbalanced sample — it exists to cover diverse agent shapes, not to model a traffic distribution, so weighting a domain by however many probe-able facts its fixtures happen to contain has no meaning to recover. (If the corpus were ever meant to mirror real traffic, micro would be the honest choice and this reasoning would invert.)

## What's reported now

Per-domain rows, which show what no single number can — the spread is real:

```
by domain (each counts once toward the macro average):
  ops/sre        100.0%     78.9%      0      38
  coding         100.0%    100.0%      0      38
  support        100.0%     80.6%      0      36
  research       100.0%     92.2%      0      64
  data-analysis  100.0%     97.6%      0      83
  devops         100.0%     83.1%      0      71
  finance        100.0%     92.0%      0      87
  web-research   100.0%      4.4%      0     666
```

Micro is printed beside macro rather than dropped, because **the two diverging is itself information** — it says one domain dominates the fact count. The JSON payload leads with `macro` and tags `micro` with the reason it moves, so a consumer can't reach for the swingable number by habit. A single-domain report keeps the old plain sentence, since a macro average over one domain adds nothing.

`--max-lost` is untouched: a lost fact is a count, not a ratio, and no averaging choice changes whether something was unrecoverable.

## Tests

The distinction is pinned rather than assumed — `test_macro_average_is_not_swung_by_one_dominant_domain` builds two 10-fact domains at 0% gap plus one 1000-fact domain at 100%, and asserts micro exceeds 0.97 while macro lands on exactly 1/3. Plus empty-domain handling, the empty report, payload ordering, and that the corpus's own macro and micro actually differ (otherwise the distinction would be untested).

lint clean · `bench`/`verify`/`validate` PASS · **2170 tests**, coverage **95.36%** · `retention.py` still 100%.

README, `docs/EVALUATION.md` and ADR 0005 §10 carry both figures and why they differ, so neither can be quoted bare.


dshakes:
## Cross-audit observation addressed (`be23315`)

Non-blocking, but valid — and it lands on code this PR added.

`{x:<14}` pads but never truncates, so any label longer than the column shoves every later column right. Two ways in, and the audit only spotted the first:

1. A custom corpus with a domain like `vision/ui-automation` (20 chars).
2. **This PR's own fallback label, `(uncategorised)` — 15 characters.** It overflowed by construction on any block without a domain. That one's mine.

Labels are now clamped to the column and the fallback is `(no domain)`:

```
'coding'                 -> 'coding'          (6)
'exactly14chars'         -> 'exactly14chars'  (14, boundary: untouched)
'vision/ui-automation'   -> 'vision/ui-aut+'  (14)
```

The clamp marker is a plain `+` deliberately rather than `…`: U+2026 has no cp1252 code point and would resurrect the Windows `UnicodeEncodeError` that took this same gate down two commits ago. There's an assertion pinning the marker is cp1252-encodable so that can't drift.

### A note on the test, since it took three attempts

The first two versions failed against output that was **correctly aligned the whole time** — I was selecting rows by "contains a %", which catches the savings-bucket labels (`50%-75%`, whose *label* has one) and the closing prose line. The third failed on `rendered.encode("cp1252")`, which is wrong because the report legitimately carries U+2588 bars under a UTF-8 console — that path is already covered properly by `test_report_encodes_on_a_windows_console`, which patches stdout.

It now asserts the property directly: `_row` emits `"  " + 14-wide label + right-aligned recall`, so the numeric column begins at index 16. Worth recording because each failure looked like a code bug and was a test bug.

### Gates

lint clean · `bench`/`verify`/`validate` PASS · **2172 tests**, coverage **95.36%** · `retention.py` still 100%.

`audit` (Codex) stays red on `Quota exceeded` — sixth consecutive PR, still exiting before it reads the diff.

PR #67, opened 2026-07-31
release(1.39.0): certify the HTML transform, and a headline one fixture cannot swing

dshakes:
Version bump for the three commits merged since `v1.38.0` — #64, #65, #66.

## Soak-exempt, and I checked rather than assumed

`RELEASING.md` exempts changes touching "nothing a running proxy/wrap/gateway executes". That needed proving here rather than asserting, because `distil/retention.py` **is** imported by the request path (the live meter) and it's in the diff.

Every request-path symbol is byte-identical to `v1.38.0`:

```
LiveMeter              IDENTICAL
measure_live           IDENTICAL
_tool_result_texts     IDENTICAL
_flatten_text          IDENTICAL
_survives              IDENTICAL
_normalize             IDENTICAL
_in_recovery           IDENTICAL
extract_targets        IDENTICAL
Tally                  IDENTICAL
_live_path             IDENTICAL

VERDICT: request path UNCHANGED — soak-exempt
```

(git's hunk headers *name* `_survives` and `format_live`, which is misleading — that's the enclosing function, not a change inside it. Hence comparing parsed function bodies rather than reading the diff.)

Only reporting, corpus fixtures and packaging moved. So this goes straight to a final, unlike 1.38.0 which was a runtime-behaviour change whose soak was skipped by decision.

## What ships

**The HTML transform is certified.** 1.38.0 shipped it default-on while admitting the corpus had no HTML. `corpus/web-research.json` fixes that — 89.8% savings, and an extractor patched to swallow `<article>` drops `match_rate` to 0.0, so the fixture can fail.

**The headline is a per-domain macro average — 21.4%.** Adding that trajectory had moved the fact-weighted figure 9.8% → 62.6% with nothing about the compressor changing. Macro counts each domain once; micro is printed beside it because the two diverging says the corpus is unbalanced. Per-domain rows show the spread: 0.0% on `coding`, 95.6% on `web-research`.

**Homebrew formula** pinned to the v1.38.0 tarball sha.

## Version surfaces — all six

| surface | 1.38.1 |
|---|---|
| `pyproject.toml` | ✅ |
| `CITATION.cff` | ✅ — moves this time; a final is citable, unlike the rc before 1.38.0. Date stays 2026-07-30 (same day) |
| `server.json` (spec + package) | ✅ ✅ |
| `packaging/npm/package.json` | ✅ |
| `plugins/distil/.claude-plugin/plugin.json` | ✅ |

Verified against `release.sh`'s own non-rc cross-surface logic: `pyproject=1.38.1 citation=1.38.1 plugin=1.38.1 server=1.38.1 npm=1.38.1`.

## One thing worth your eye

This is a **patch** number carrying *additive* `distil retention --json` keys (`macro`, `micro`, `by_domain`) and a new corpus domain. Nothing breaks, so patch is defensible — but a strict SemVer reading makes new backwards-compatible functionality a **minor**, i.e. 1.39.0. Your call; I've cut what you asked for and noted it in the commit so the history isn't silent about it.

## Gates

lint + `format --check` clean · `bench`/`verify`/`validate` PASS · **2172 tests**, coverage **95.36%** · packaging smoke 14/14.

After merge: `./scripts/release.sh` from `main` pushes `v1.38.1`; CI publishes to PyPI, builds the image, bumps the tap.


dshakes:
**Retitled 1.38.1 → 1.39.0** (`c68ea0f`), per review.

The changes are additive rather than corrective: `distil retention --json` gained `macro`, `micro` and `by_domain` keys, the text report gained per-domain rows, and the corpus gained a domain. Nothing breaks — so a patch would not have been *wrong* — but SemVer calls new backwards-compatible functionality a minor, and a version number that understates what moved is a small lie told to whoever reads the version histogram later.

All six surfaces move together: `pyproject`, `CITATION.cff`, `server.json` (×2), the npm wrapper, the Claude Code plugin. Re-verified against `release.sh`'s non-rc cross-surface logic — all five agree on `1.39.0` — and packaging smoke is 14/14.

The soak-exemption evidence in the description is unchanged and still holds: every request-path symbol is byte-identical to `v1.38.0`.

PR #69, opened 2026-07-31
fix(release): a missing npm token must fail the release, not pass it

dshakes:
`publish-npm` exited 0 when `NPM_TOKEN` was absent, warning into a **green** run.

## What that cost

The registry sat on **1.25.0 while PyPI reached 1.39.0** — fourteen releases, each reporting success, publishing nothing. npm itself is the evidence:

```
versions on registry: ['1.25.0', '1.39.0']
```

Every release in between was skipped without anyone being told, while the README carried a live npm badge and an `npm i distil-llm` instruction aimed at the stale build.

**A warning on a green job is not a signal.** Nobody reads the log of a step that passed — which is exactly why this survived fourteen chances to be noticed.

## The change

A missing token now fails, and the error says what to do — including the two things that actually bit while setting this up today:

```
::error::NPM_TOKEN is not set, so this release cannot publish to npm.
::error::Failing rather than skipping: a silent skip here left the
::error::registry fourteen releases behind while every run stayed green.
::error::Add an npm *automation* token as the NPM_TOKEN repo secret
::error::(a Publish/classic token still triggers 2FA and fails EOTP),
::error::and ensure the package's Publishing access allows tokens.
```

Precedent is `publish-pypi`: dormant behind an explicit opt-in, then failing for real. Quietly succeeding isn't one of its states either.

**Forks** are excluded at the `if:` (`github.repository == 'dshakes/distil'`) rather than inside the step, so a fork that tags a release never runs the job and never sees a failure it can't fix.

**Idempotency is untouched** — re-running a release for a version already on npm still exits 0.

## Verified by running it, not reasoning about it

Both branches were exercised against the extracted step body:

| case | result |
|---|---|
| `NODE_AUTH_TOKEN=""` | **exit 1** + the guidance above |
| token set, version already published | `distil-llm@1.39.0 already on npm — nothing to do`, exit 0 |

(I checked the exit code directly rather than through a pipe — a `tail`/`PIPESTATUS` reading is what gave me a wrong status earlier in this same npm investigation.)

YAML parses; packaging smoke 14/14; lint clean.

## Not addressed here

`bump-homebrew` has the same `exit 0` on a missing `TAP_TOKEN`. It currently works because the token is set, so I've left it alone rather than widen this PR — but it is the identical trap, and worth the same treatment if you want it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -197,21 +197,27 @@ jobs:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: |
           set -euo pipefail
+          cd packaging/npm
+          V="$(node -p "require('./package.json').version")"
+          # Idempotency first, and deliberately BEFORE the token gate: re-cutting a
+          # release for a version already on npm has nothing to publish, so it must
+          # neither fail nor require a credential. `npm view` reads a public package
+          # without auth. (Audit finding on #69 — the token gate as first written
+          # would have failed a re-run that had no work to do.)
+          if npm view "distil-llm@$V" version >/dev/null 2>&1; then
+            echo "distil-llm@$V already on npm — nothing to do"
+            exit 0
+          fi
+          # Past here there IS something to publish, so a missing token is a real
+          # failure rather than a shrug.
           if [ -z "${NODE_AUTH_TOKEN:-}" ]; then
-            echo "::error::NPM_TOKEN is not set, so this release cannot publish to npm."
+            echo "::error::NPM_TOKEN is not set, so distil-llm@$V cannot be published."
             echo "::error::Failing rather than skipping: a silent skip here left the"
             echo "::error::registry fourteen releases behind while every run stayed green."
             echo "::error::Add an npm *automation* token as the NPM_TOKEN repo secret"
             echo "::error::(a Publish/classic token still triggers 2FA and fails EOTP),"
             echo "::error::and ensure the package's Publishing access allows tokens."
             exit 1
-          fi
-          cd packaging/npm
-          V="$(node -p "require('./package.json').version")"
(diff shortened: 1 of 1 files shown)


dshakes:
## Codex finding addressed (`f6b2fe1`) — and it was right

> **Medium:** `publish-npm` now fails before checking whether the npm version already exists […] a rerun/re-cut for an already-published version fails even though the workflow explicitly says rerunning an existing npm version "must not fail".

Correct, and a real contradiction I introduced: the token gate sat ahead of the idempotency check, so a re-cut with nothing to publish would fail for want of a credential it had no use for — against the intent stated two steps below it.

Adopted the recommendation exactly. Idempotency now runs first; `npm view` reads a public package without auth, so the no-work case exits 0 regardless of the token. Past that point there genuinely *is* something to publish, which is precisely where a missing token should be fatal. The loud failure survives where it matters and disappears where it was noise.

### Verified per branch, by running it

| case | expected | actual |
|---|---|---|
| **no token**, 1.39.0 already on npm ← the finding's case | exit 0 | `exit=0` — `distil-llm@1.39.0 already on npm — nothing to do` |
| token set, already on npm | exit 0 | `exit=0` — same |
| **no token**, version *not* on npm | exit 1 | `exit=1` — `::error::…Publishing access allows tokens` |

The first row also confirms the load-bearing assumption in the fix: `npm view` really does succeed with an empty `NODE_AUTH_TOKEN`, so moving the check earlier doesn't just relocate the failure.

---

Worth noting this is the **first substantive finding Codex has produced since the quota was restored** — eight PRs (#61–#68) merged with only the Gemini audit, and this one caught a logic ordering bug that the Gemini pass on the same diff did not. Good argument for keeping both.

The `bump-homebrew` `exit 0`-on-missing-token note in the description still stands as untouched follow-up.

PR #71, opened 2026-07-31
ci(audit): audit the fix, not just the code that needed fixing

dshakes:
The Codex auditor triggered on `opened, reopened, labeled` — **not `synchronize`** — so it only ever saw the commit that *contained* a bug. It reviewed, a fix was pushed, and the fix shipped unaudited unless someone remembered the `agent:audit` label.

That's backwards: a correction to a Blocking finding deserves at least the scrutiny of the code that caused it. It bit twice today — on #69 and #70 Codex found real bugs in release-workflow logic, and in both cases my correction went in unreviewed until I applied the label by hand.

## The exclusion was deliberate, and its reason still holds

```
# Not on `synchronize`, so the fix loop's pushes don't re-spend Codex credits each round.
```

That is not a hypothetical cost. The quota ran dry earlier today and **eight PRs (#61–#68) merged with one auditor instead of two.** So this doesn't just delete the guard — it moves the cost filter to where the cost actually lives.

`synchronize` is added to the trigger, and **bot actors are excluded in the job condition**:

```yaml
if: >-
  !endsWith(github.actor, '[bot]')
  && (github.event.action != 'labeled' || github.event.label.name == 'agent:audit')
```

Human pushes — the ones carrying fixes a reviewer hasn't seen — are audited. The auto-fix loop's per-round commits are not, because the loop is already gated by the reviewer and a second opinion per round buys nothing. This also subsumes the old `dependabot[bot]` carve-out, which existed because those runs get Dependabot's empty secret store and can't auth at all.

## Verified across every combination

Reasoning about GitHub expression logic is how you get it subtly wrong, so the condition was evaluated directly:

| actor | event | runs | why |
|---|---|---|---|
| human | `opened` | ✅ | unchanged |
| human | **`synchronize`** | ✅ | **the gap this closes** |
| `compass-agent[bot]` | `synchronize` | ❌ | **credits preserved** |
| `github-actions[bot]` | `synchronize` | ❌ | workflow-authored push |
| `dependabot[bot]` | `opened` | ❌ | empty secret store |
| human | `labeled` `agent:audit` | ✅ | manual re-audit |
| human | `labeled` `agent:qa` | ❌ | unrelated label |
| human | `reopened` | ✅ | unchanged |

## One residual, recorded in the workflow rather than left to be found

The fix loop falls back to `SDLC_BOT_TOKEN` when `SDLC_APP_ID` is unset, and **a PAT pushes under its owner's name, not a `[bot]` actor** — so in that configuration the loop's rounds *will* be audited and will spend credits. The App path is the documented one. Flagged in the workflow comments so it's a known trade rather than a surprise bill.

## Not included

`sdlc-audit-gemini.yml` has the identical trigger and therefore the identical gap. I've left it alone — it's a different provider and quota, so the cost calculus isn't the same one you just weighed. Say the word and it's a one-line follow-up.

### DIFF
--- a/.github/workflows/sdlc-audit.yml
+++ b/.github/workflows/sdlc-audit.yml
@@ -18,25 +18,41 @@ on:
   pull_request:
     types: [opened, reopened, synchronize, labeled]
 
+# One audit per PR. `synchronize` makes rapid pushes possible, and without this an
+# older run can finish LAST and post a verdict — clean or blocking — against a commit
+# that has already been superseded. Cancelling in flight also stops a push storm from
+# spending Codex credits on commits nobody will merge. (Audit finding on #71.)
+concurrency:
+  group: sdlc-audit-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
   pull-requests: write
 
 jobs:
   audit:
-    # Auto on open/reopen and on human pushes, or on the agent:audit label. (label.name is
-    # null on open events.) Bot actors are excluded for two separate reasons that happen to
-    # share a filter: dependabot runs get Dependabot's (empty) secret store so the agent
-    # cannot auth at all, and the SDLC fix loop pushes a commit per round — auditing every
-    # one of those re-spends Codex credits for no added signal, which is why `synchronize`
-    # was originally left out of the trigger entirely.
+    # Three independent guards, each for its own reason:
     #
-    # Residual: the loop falls back to SDLC_BOT_TOKEN when SDLC_APP_ID is unset, and a PAT
-    # pushes under its owner's name rather than a `[bot]` actor. In that configuration the
-    # loop's rounds WILL be audited. The App path is the documented one; this is noted so
-    # the cost is a known trade rather than a surprise.
+    # 1. dependabot never audits — those runs get Dependabot's (empty) secret store, so the
+    #    agent cannot auth at all. This is about capability, not cost.
+    # 2. On `synchronize` only, bot pushes are skipped: the SDLC fix loop commits once per
+    #    round and auditing each re-spends Codex credits for no added signal, since the loop
+    #    is already gated by the reviewer. Scoped to synchronize deliberately — a blanket
+    #    bot exclusion would also stop auditing PRs *opened* by Renovate, Copilot, release
+    #    bots and the like, which are dependency and generated-code changes and among the
+    #    most worth a second opinion. (Audit finding on #71: the first version of this
+    #    filter made exactly that mistake.)
+    # 3. A label only triggers an audit when it is `agent:audit`. (label.name is null on
+    #    open events, hence the action check first.)
+    #
+    # Residual: the fix loop falls back to SDLC_BOT_TOKEN when SDLC_APP_ID is unset, and a
+    # PAT pushes under its owner's name rather than a `[bot]` actor, so in that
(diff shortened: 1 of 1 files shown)


dshakes:
## Both findings addressed (`0864fcb`) — both correct, both self-inflicted

### Blocking: bot *pushes* vs bot *pull requests*

> changing the guard from only `dependabot[bot]` to `!endsWith(github.actor, '[bot]')` disables automatic Codex audits for every non-Dependabot bot PR too, including Renovate, Copilot/Claude app PRs, release bots

Right, and it's the more serious of the two. I was trying to spare the fix loop's per-round pushes and instead withdrew coverage from a whole class of PR — dependency bumps and generated code, which are among the changes a second opinion is *most* worth on. Saving credits on one event quietly cost auditing on another.

Adopted the recommendation exactly: the skip is now scoped to `synchronize`, the only event the fix loop produces.

```yaml
github.actor != 'dependabot[bot]'
&& (github.event.action != 'synchronize' || !endsWith(github.actor, '[bot]'))
&& (github.event.action != 'labeled' || github.event.label.name == 'agent:audit')
```

dependabot keeps its own unconditional exclusion, which was never about cost — those runs get Dependabot's empty secret store, so the agent can't authenticate at all. Worth separating the two reasons that had been sharing one filter.

### Should Fix: concurrency

> rapid pushes can leave multiple Codex audits running; an older audit can finish last and post a stale "clean" or blocking comment against a superseded commit

Also right, and a race that only becomes possible *because* I added `synchronize`. A stale "no findings" on a PR that has moved on is worse than no audit at all, because it reads as coverage.

```yaml
concurrency:
  group: sdlc-audit-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

Per-PR, cancel in flight — which also stops a push storm from spending credits on commits nobody will merge, so it serves the cost goal too.

### Re-verified, including the cases the finding named

| actor | event | runs |
|---|---|---|
| human | `opened` / `reopened` / `synchronize` | ✅ |
| **`renovate[bot]`** | `opened` | ✅ **restored** |
| **`copilot[bot]`** | `opened` | ✅ **restored** |
| `compass-agent[bot]` | `opened` | ✅ |
| `compass-agent[bot]` | `synchronize` | ❌ fix-loop round |
| `github-actions[bot]` | `synchronize` | ❌ |
| `dependabot[bot]` | any | ❌ empty secret store |
| human | `labeled` `agent:audit` / other | ✅ / ❌ |

Twelve combinations evaluated against the condition rather than reasoned about.

---

Third PR running where this auditor caught a real bug in my work — #69's ordering, #70's version-only idempotency check, and now this. All three were changes I had tested; the tests passed because they covered the cases I'd thought of. That is the argument for the second opinion, and it's also the argument this PR is making.

dshakes:
**Residual closed.** The audit noted it couldn't run `actionlint` in its environment, so the review was expression-only. Ran it here:

```console
$ actionlint .github/workflows/sdlc-audit.yml
$ echo $?
0
```

Clean — no schema, expression, or shell issues in the workflow.

Also worth recording: **this audit fired automatically on the fix push, not from a label.** The previous one ran on `dce3582` at 05:19 and this one on `0864fcb` at 05:22, with no label added and no reopen in between. Under the old trigger that push would have produced nothing — so `synchronize` did the work. The change validated itself on the very commit that fixes the findings against it, which is the behaviour the PR is arguing for.

PR #72, opened 2026-07-31
ci(audit): audit the fix in Gemini too, with its own concurrency group

dshakes:
#71 fixed this for the Codex auditor. Gemini carried the identical gap: it triggered on `opened, reopened, labeled` but not `synchronize`, so it reviewed the commit that *contained* a bug and never the fix, unless someone applied `agent:audit` by hand.

Two auditors sharing one blind spot isn't two opinions on the fix.

## Same shape as #71, including the mistake it already paid for

Bot pushes are skipped **only on `synchronize`** — the sole event the SDLC fix loop produces — so its per-round commits don't re-spend Gemini credits. Bot-*opened* PRs (Renovate, Copilot, release bots) keep auditing.

That distinction isn't incidental: a blanket `!endsWith(actor,'[bot]')` was filed **Blocking** on #71 precisely because it withdraws coverage from dependency and generated-code changes. Applying the corrected form here rather than rediscovering it.

`dependabot` keeps its unconditional skip, which is about *capability* not cost — those runs get an empty secret store and can't authenticate.

## The detail worth the most care: the concurrency group

```yaml
concurrency:
  group: sdlc-audit-gemini-${{ github.event.pull_request.number }}   # NOT sdlc-audit-…
  cancel-in-progress: true
```

**Concurrency groups are shared across workflows.** Reusing `sdlc-audit-${{ … }}` would make the two auditors cancel each other, quietly reducing three opinions to whichever started last — exactly the failure this project keeps finding: a green run that did less than it appears to have.

Asserted distinct in the same commit:

```
gemini group: sdlc-audit-gemini-${{ github.event.pull_request.number }}
codex  group: sdlc-audit-${{ github.event.pull_request.number }}
groups are distinct: auditors cannot cancel each other
```

## The injection defence is untouched

This file's security property is that it checks out the **base** commit and never places the PR tree on disk, so nothing the Gemini CLI auto-loads from a workspace — `GEMINI.md` as agent context, `.env`, `.gemini/settings.json` whose `mcpServers` it spawns as processes — can come from a contributor. The PR enters as data only.

Verified by diffing specifically for `ref:`/`checkout`/`base`/`GEMINI.md`/`mcpServers`/`.env` lines rather than reading past them: **no change to any of them.** Only the trigger, the concurrency block, and the job condition moved.

## Verified

- 9 event/actor combinations evaluated against the condition (human push runs, `compass-agent[bot]` push skips, `renovate[bot]`/`copilot[bot]` opens still run, dependabot never)
- `actionlint` clean
- Concurrency groups asserted distinct

## Note, not addressed here

Gemini "skips cleanly (no failure) when `GEMINI_API_KEY` is absent" — the same green-job-that-did-nothing shape #69 and #70 removed from the publish jobs. The consequence differs (a lost internal signal, not a user-facing artefact that silently never shipped), so I've left it rather than widen this PR. Worth deciding on separately.

### DIFF
--- a/.github/workflows/sdlc-audit-gemini.yml
+++ b/.github/workflows/sdlc-audit-gemini.yml
@@ -21,7 +21,19 @@ on:
 # and silently reduce three opinions to whichever finished first.
 concurrency:
   group: sdlc-audit-gemini-${{ github.event.pull_request.number }}
-  cancel-in-progress: true
+  # Cancel ONLY when this run will actually audit. Workflow-level concurrency is
+  # evaluated BEFORE the job-level `if`, so a run whose job will be skipped — a bot
+  # `synchronize`, an unrelated label — still joins the group. With a bare
+  # `cancel-in-progress: true` it would kill a valid audit already in flight and then
+  # skip, leaving the PR with no verdict at all: the precise opposite of the coverage
+  # this trigger exists to add. (Audit finding on #72; #71 shipped with it.)
+  #
+  # The predicate is the job's `if`, deliberately duplicated rather than referenced —
+  # workflow-level concurrency cannot see job outputs. Keep the two in step.
+  cancel-in-progress: >-
+    ${{ github.actor != 'dependabot[bot]'
+    && (github.event.action != 'synchronize' || !endsWith(github.actor, '[bot]'))
+    && (github.event.action != 'labeled' || github.event.label.name == 'agent:audit') }}
 
 permissions:
   contents: read
--- a/.github/workflows/sdlc-audit-gemini.yml
+++ b/.github/workflows/sdlc-audit-gemini.yml
@@ -1,13 +1,27 @@
-# Auditor (Gemini) — third cross-model audit. Runs automatically when a PR opens (an
-# independent third opinion alongside the Claude reviewer and Codex auditor) and on
-# demand via the `agent:audit` label. Read-only; posts Gemini's verdict as a PR comment.
-# Needs secret GEMINI_API_KEY — skips cleanly (no failure) when it is absent.
-# Not on `synchronize`, so the fix loop's pushes don't re-spend Gemini credits each round.
+# Auditor (Gemini) — third cross-model audit. Runs when a PR opens, when a human pushes
+# to it, and on demand via the `agent:audit` label. An independent third opinion
+# alongside the Claude reviewer and the Codex auditor. Read-only; posts Gemini's verdict
+# as a PR comment. Needs secret GEMINI_API_KEY — skips cleanly (no failure) when absent.
+#
+# `synchronize` is in the list for the reason given in sdlc-audit.yml (#71): without it
+# the auditor only ever saw the commit that CONTAINED a bug, and the fix shipped
+# unreviewed unless someone remembered the label. The cost that originally justified
+# leaving it out is filtered in the job condition instead — see there.
 name: "sdlc · audit (Gemini)"
 
 on:
   pull_request:
-    types: [opened, reopened, labeled]
+    types: [opened, reopened, synchronize, labeled]
+
+# One Gemini audit per PR, so a rapid push cannot leave an older run to finish last and
+# post a verdict against a superseded commit.
+#
+# The group name is deliberately NOT the one sdlc-audit.yml uses: concurrency groups are
+# shared across workflows, so a common name would make the two auditors cancel each other
+# and silently reduce three opinions to whichever finished first.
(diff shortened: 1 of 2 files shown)


dshakes:
## Blocking finding addressed (`a45ae8b`) — and it reaches back into #71

> Workflow-level `concurrency` applies before the job-level `if`, so skipped events can still cancel an in-progress valid audit […] That can leave the PR with no Gemini verdict, which is the opposite of the new `synchronize` coverage goal.

Correct, and the nastiest failure in this whole sequence. Concretely:

1. A human pushes a fix → audit starts, running
2. The fix loop pushes 10s later → `synchronize` → joins the same group → `cancel-in-progress: true` **kills the running audit**
3. The job then skips on its `if`

Net result: **no verdict at all** — worse than the gap it replaced, because a missing audit looks exactly like a completed one. And since `#71` shipped the identical pattern for Codex, this fixes **both files**.

### The fix

`cancel-in-progress` is now the job's own predicate, so a run can only cancel an in-flight audit if it will produce one itself:

```yaml
cancel-in-progress: >-
  ${{ github.actor != 'dependabot[bot]'
      && (github.event.action != 'synchronize' || !endsWith(github.actor, '[bot]'))
      && (github.event.action != 'labeled' || github.event.label.name == 'agent:audit') }}
```

Skipped events still join the group but queue harmlessly behind the running audit and then skip.

**Why not job-level concurrency**, which the finding offered first: it was hedged with *"if skipped jobs do not acquire it"*, and that depends on scheduler semantics I can't verify from here. An expression on `cancel-in-progress` is documented and deterministic, so it doesn't rest on an assumption I'd be unable to test. The predicate is duplicated rather than referenced because workflow-level concurrency can't see job outputs; both copies carry a note to keep them in step.

### Verified — the invariant is "cancels ⟹ runs"

| actor | event | cancels | runs |
|---|---|---|---|
| human | `synchronize` | ✅ | ✅ supersedes correctly |
| `compass-agent[bot]` | `synchronize` | ❌ | ❌ **can't kill a live audit** |
| human | `labeled` `agent:qa` | ❌ | ❌ **can't kill a live audit** |
| `dependabot[bot]` | `synchronize` | ❌ | ❌ |
| human | `labeled` `agent:audit` | ✅ | ✅ |
| human | `opened` | ✅ | ✅ |

Asserted, plus `actionlint` clean on both workflows.

---

Fourth consecutive PR where this auditor caught something real, and the first to catch a bug in **already-merged** code. Adding `synchronize` is what made these races reachable at all — the coverage improvement created the hazard, and the auditor found it before it could bite.

dshakes:
## Second Blocking finding addressed (`ab0ca87`) — and it was right about my fix, not just my bug

> The conditional `cancel-in-progress` only protects the currently running audit; GitHub Actions still keeps only one pending run by default, and a newly queued run replaces any existing pending run in the same group.

Correct. My guard stopped a skipped event from cancelling a **running** audit, but the run still joins a workflow-level group — and with one pending slot per group, a newly queued run **evicts the one waiting**. An unrelated label or bot push could still displace a queued real audit and then skip, leaving the latest human push unaudited. I closed half the hole and reported it as closed.

### Taking the recommendation I declined last round

Concurrency now sits on the **job**:

```yaml
  audit:
    if: >-
      github.actor != 'dependabot[bot]' && …
    runs-on: ubuntu-latest
    concurrency:
      group: sdlc-audit-${{ github.event.pull_request.number }}
      cancel-in-progress: true
```

A job skipped by its `if` never joins the group at all, so no event this workflow ignores can cancel *or* evict anything. The class is gone by construction rather than by an expression that has to enumerate every skip case — and it removes the duplicated predicate, so there's one condition to maintain instead of two that must agree.

### On why I got this wrong the first time

I declined this exact suggestion because it was hedged with *"if skipped jobs do not acquire it"* and I couldn't verify the scheduler from here. But I then kept a construct I **also** couldn't verify — one whose correctness depended on pending-slot semantics I'd never tested — and that turned out to be the one with the hole. Given equal uncertainty, the simpler mechanism was the better bet, and "I can't test it" argued *for* the recommendation rather than against it.

### Unchanged and re-verified

- Same triggers, same three guards on both workflows
- Group names stay per-workflow — the auditors still can't cancel each other
- Gemini still checks out the **base** commit, PR tree never on disk (diffed for `ref`/`checkout`/`base`/`GEMINI.md`/`mcpServers`/`.env`: none touched)
- No workflow-level `concurrency` remains in either file
- `actionlint` clean on both

PR #75, opened 2026-07-31
ci(audit): a missing Gemini key is a misconfiguration, not a quiet no-op

dshakes:
The last green-job-that-did-nothing in the repo. `GEMINI_API_KEY` absent made the job `exit 0` with an informational line — **indistinguishable from an audit that ran and found nothing.** That's the exact shape that let npm sit fourteen releases behind a green publish job until #69.

## Why this one couldn't take the npm treatment as-is

**GitHub deliberately withholds secrets from `pull_request` runs raised by a fork.** A blanket "fail when the key is absent" would therefore redden *every external contributor's PR* to enforce our own configuration — a self-inflicted wound, not a signal. Pattern-matching #69/#70 here would have shipped that.

So the check now distinguishes the two cases it had been collapsing:

```yaml
SAME_REPO: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
```

| situation | before | now |
|---|---|---|
| key present | audit runs | audit runs |
| **absent, same-repo PR** | silent green | **fails**, with what to fix |
| **absent, fork PR** | silent green | skips green, says why |

On a fork the notice also points out that Codex and the Claude reviewer still cover the PR — the *third* opinion is missing, not all of them.

## Verified by running the step, four ways

| key | same-repo | exit | output |
|---|---|---|---|
| present | true | 0 | `present=true` |
| **absent** | **true** | **1** | `::error::…Add GEMINI_API_KEY as a repo secret` |
| absent | false | 0 | `present=false`, `::notice::fork PR — secrets withheld by design` |
| present | false | 0 | `present=true` |

The third row matters most: `present=false` is still emitted, so the later steps stay gated exactly as they were — the fork path is byte-for-byte the old behaviour.

## Unchanged

The injection defence is untouched: the job still checks out the **base** commit with the PR tree never on disk. Verified by diffing specifically for `ref`/`checkout`/`base`/`GEMINI.md`/`mcpServers`/`.env` — none touched. `actionlint` clean.

## After this

Every job in the repo that can silently do nothing now either fails or says why:

| job | state |
|---|---|
| `publish-npm` | fails on missing token (#69) |
| `bump-homebrew` | fails on missing token (#70) |
| `publish-pypi` | dormant behind an opt-in, then fails for real |
| `publish-image` | needs no secret |
| `audit` (Codex) | needs `OPENAI_API_KEY`; the action fails outright when absent |
| `audit-gemini` | **this PR** |

### DIFF
--- a/.github/workflows/sdlc-audit-gemini.yml
+++ b/.github/workflows/sdlc-audit-gemini.yml
@@ -55,13 +55,34 @@ jobs:
         id: key
         env:
           GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+          # GitHub withholds secrets from `pull_request` runs raised by a fork, so an
+          # absent key there is by design, not misconfiguration. Only a same-repo PR can
+          # tell the two apart — and failing every external contributor's PR to enforce
+          # our own config would be a self-inflicted wound.
+          SAME_REPO: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
         run: |
+          set -euo pipefail
           if [ -n "$GEMINI_API_KEY" ]; then
             echo "present=true" >>"$GITHUB_OUTPUT"
-          else
-            echo "present=false" >>"$GITHUB_OUTPUT"
-            echo "GEMINI_API_KEY not set — skipping the Gemini cross-audit."
+            exit 0
           fi
+          if [ "$SAME_REPO" != "true" ]; then
+            echo "present=false" >>"$GITHUB_OUTPUT"
+            echo "::notice::fork PR — GitHub withholds secrets from these runs, so the"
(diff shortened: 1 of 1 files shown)


dshakes:
**Residual fixed** (`65731c8`).

> the file header still says the workflow "skips cleanly (no failure) when absent," which is now stale for same-repo PRs

Correct — and the same commit that made it false left it in place. A comment contradicting the code beneath it is worse than no comment, since it's what the next reader trusts before reading the shell.

The header now states both cases:

```
# Needs secret GEMINI_API_KEY: absent on a same-repo PR that is a
# misconfiguration and fails; absent on a fork PR it skips green, because GitHub
# withholds secrets from those runs by design.
```

`actionlint` still clean.

PR #76, opened 2026-07-31
fix(sdlc): stop claiming the reviewer will re-run when it cannot

dshakes:
## I can't make the loop close — but I can stop it lying about being closed

**The credential is the fix, and only you can add it.** What I can fix is that the loop currently *asserts* something false.

## The finding

`sdlc-fix` pushes with `BOT_TOKEN`, which falls through to `github.token`:

```yaml
BOT_TOKEN: ${{ steps.apptoken.outputs.token || secrets.SDLC_BOT_TOKEN || github.token }}
```

And that fallback is what's live — confirmed against the repo config:

```console
$ gh api repos/dshakes/distil/actions/variables --jq '.variables[].name'
PUBLISH_TO_PYPI                        # no SDLC_APP_ID

$ gh api repos/dshakes/distil/actions/secrets --jq '.secrets[].name'
ANTHROPIC_API_KEY  CLAUDE_CODE_OAUTH_TOKEN  GEMINI_API_KEY
NPM_TOKEN  OPENAI_API_KEY  TAP_TOKEN  TRAFFIC_TOKEN     # no SDLC_BOT_TOKEN
```

**GitHub raises no workflow events for pushes authenticated with `GITHUB_TOKEN`** — a deliberate recursion guard. So an auto-fix push re-runs *nothing*: not the reviewer, not the cross-audits, not the gate matrix. The fix lands unvalidated, `agent:needs-fix` stays, the loop halts.

Meanwhile the job printed:

```
Pushed fix to $HEAD_REF — Reviewer will re-run.
```

…and a comment two steps below asserted the same. A green job stating something false is precisely what #69, #70 and #75 removed from the publish jobs — here it's in the loop that's meant to *fix* things.

## The change

Truthful in both states. With a bot credential: the existing message. Without one: a warning **and a PR comment** saying the fix is unvalidated, nothing re-ran, the label is still set, and what to configure.

The comment matters more than the annotation — nobody reads the log of a job that passed, but the PR is where the human already is.

## Deliberately does *not* fail the job

`sdlc-ci-fix` triggers on `check_suite: completed`. A new failure here would feed the CI medic, which relabels `agent:needs-fix` and re-triggers this very workflow. **Turning a silent stall into a loop would be a worse bug than the one being fixed** — so this warns loudly instead, which is the opposite of the call made on the publish jobs and for a specific reason.

## To actually close the loop

| option | effect |
|---|---|
| `SDLC_APP_ID` + `SDLC_APP_PRIVATE_KEY` | documented path; App pushes trigger workflows |
| `SDLC_BOT_TOKEN` (PAT) | fallback; also triggers |

Either makes `HAS_BOT_CRED` true and the old message accurate again — no further code change needed.

## Verified

Both branches exercised, shell parses, `actionlint` clean.

Two self-inflicted problems caught on the way: my first version broke the YAML (a multi-line `--body` escaped the block scalar), and my `actionlint exit=$?` was reading `head`'s status through a pipe — the same trap that produced a wrong readout twice earlier today. Both fixed; the SC2016 suppression carries its reason, since the backticks are markdown code spans rather than shell.

### DIFF
--- a/.github/workflows/sdlc-fix.yml
+++ b/.github/workflows/sdlc-fix.yml
@@ -174,18 +174,24 @@ jobs:
               # green job asserting something false, which is the failure this repo
               # keeps finding.
               echo "::warning::Fix pushed to $HEAD_REF, but NOTHING will re-validate it."
-              echo "::warning::Pushes made with GITHUB_TOKEN do not trigger workflow runs,"
-              echo "::warning::so no reviewer, audit or gate will see this commit and the"
-              echo "::warning::auto-fix loop stops here. Configure SDLC_APP_ID +"
-              echo "::warning::SDLC_APP_PRIVATE_KEY (preferred) or the SDLC_BOT_TOKEN secret."
+              echo "::warning::No usable bot credential was available, so the push fell back"
+              echo "::warning::to GITHUB_TOKEN — and GitHub raises no workflow events for"
+              echo "::warning::those, so no reviewer, audit or gate sees this commit."
+              echo "::warning::Either no credential is configured (SDLC_APP_ID +"
+              echo "::warning::SDLC_APP_PRIVATE_KEY, or SDLC_BOT_TOKEN), or the App token"
+              echo "::warning::step failed to mint one — it runs under continue-on-error, so"
+              echo "::warning::check its log before assuming the secrets are missing."
               # shellcheck disable=SC2016  # the backticks below are markdown code
               # spans in a PR comment, not shell expansions.
               gh pr comment "$PR" --body "$(printf '%s\n' \
                 '⚠️ **Auto-fix pushed, but nothing re-ran.**' '' \
                 "A fix was committed to \`$HEAD_REF\`, and then no checks re-ran against it." \
-                'GitHub raises no workflow events for pushes authenticated with `GITHUB_TOKEN`,' \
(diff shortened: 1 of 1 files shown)


dshakes:
**Should-fix addressed** (`6d36a09`) — and it's the kind of bug that only matters in the message that exists to prevent bugs.

> this branch also happens when `SDLC_APP_ID` is configured and `actions/create-github-app-token` fails under `continue-on-error` due to missing/invalid `SDLC_APP_PRIVATE_KEY` or App permissions. That will send maintainers toward the wrong diagnosis.

Correct. The step is `continue-on-error: true` **by design** — so a bad private key or a missing App permission doesn't abort the job — which is exactly what makes that failure invisible unless someone opens its log. My wording would have sent a maintainer hunting for secrets already sitting in the repo.

For a message whose only purpose is diagnosis, naming the wrong cause is the entire defect.

Both the annotation and the PR comment now say **no usable bot credential was available** and give both branches:

```
Either no credential is configured (SDLC_APP_ID + SDLC_APP_PRIVATE_KEY, or
SDLC_BOT_TOKEN), or the App token step failed to mint one — it runs under
continue-on-error, so check its log before assuming the secrets are missing.
```

YAML parses, shell parses, `actionlint` clean.

PR #78, opened 2026-08-03
test(sdlc): prove the auto-fix loop closes [do not merge]

dshakes:
## Do not merge — throwaway

This PR exists to prove one thing: that the SDLC auto-fix loop actually closes, now that `SDLC_APP_ID` / `SDLC_APP_PRIVATE_KEY` are wired and the App holds all eight required permissions.

That loop has **never run once** in this repo — 8 `sdlc-fix` invocations, all skipped, and until today the `agent:needs-fix` label did not exist here.

`distil/_looptest/window.py` is deliberately defective:
- `tail_window` has an off-by-one (`[-1]` drops the final element, so the window is always one short)
- `mean_of_tail` divides by `size` with no guard for `size == 0`

### What should happen

1. **Reviewer** returns BLOCKING and applies `agent:needs-fix` — using the App token (`sdlc-review.yml:124`), so the actor is `compass-sdlc-bot[bot]`, not `github-actions[bot]`.
2. **`sdlc-fix`** fires. The new **Authorize the trigger actor** gate should allow it via path 1 and log *"is the App that minted this run's token"* — path 2 cannot help, since `/apps/compass-sdlc-bot` 404s (private App).
3. Builder fixes both defects and pushes.
4. **That push re-triggers the Reviewer.** This is the whole point — a `GITHUB_TOKEN` push raises no workflow events, so before today this step simply did not happen.

Step 4 is the proof. Everything before it has been verified in isolation; only a live run shows the chain.

Close and delete the branch once observed. Nothing here should reach `main`.

### DIFF
--- a/distil/_looptest/window.py
+++ b/distil/_looptest/window.py
@@ -1,7 +1,7 @@
 """Scratch module used to prove the SDLC auto-fix loop closes end to end.
 
-This file is deliberately defective. It exists on a throwaway branch to trigger a
-real Reviewer -> agent:needs-fix -> Builder -> push -> re-review cycle. Delete the
+This file exists on a throwaway branch to exercise a real
+Reviewer -> agent:needs-fix -> Builder -> push -> re-review cycle. Delete the
 branch afterwards; it must never reach main.
 """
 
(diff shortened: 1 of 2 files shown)


dshakes:
Proof complete — closing unmerged, as intended. Nothing here reaches `main`.

**Round 1** (with `SDLC_TRIGGER_BOTS` set):
`c19dcf2` BLOCKING → label by `compass-sdlc-bot[bot]` → fix → push `99f3fae` → **re-review on the new SHA** → `agent:reviewed-clean`.

**Round 2** (with `SDLC_TRIGGER_BOTS` deleted) — the real test:
the gate had no hand-set allowlist to match, and `/apps/compass-sdlc-bot` 404s (private App), so path 2 could not help either. It authorised anyway:

```
Actor 'compass-sdlc-bot[bot]' is the App that minted this run's token.
```

Then `5a9a313` → fix → push `d6e6f95` → re-review. Both bot fixes were verified independently here, not taken on the reviewer.

This is the first time the auto-fix loop has closed in this repo — the previous 8 `sdlc-fix` invocations all skipped.

`SDLC_TRIGGER_BOTS` stays deleted: path 1 carries it, so the App self-authorises and there is no variable to go stale.

PR #134, opened 2026-08-15
release(1.47.0rc2): re-cut so the soak runs against a fully-gated artifact

dshakes:
Re-cut of `1.47.0rc1`. **The code is byte-identical** — nothing was wrong with it.

## Why re-cut

rc1 was validated by a preflight gate that was quietly weaker than CI. The release venv lacked `cryptography` and `opentelemetry`, so those tests `importorskip`'d themselves:

| | passed | skipped |
|---|---|---|
| gate that validated rc1 | 3381 | 6 |
| CI | 3385 | 2 |

Four tests — at-rest crypto among them — never ran against the published artifact. All four pass now, so this is not a defect fix; it is so the 3-day soak clock runs against an artifact whose gate actually matched CI, instead of one carrying a caveat. #133 added the guard that makes that state impossible to reach silently again.

## Version surfaces

All bumped to `1.47.0rc2` (npm as `1.47.0-rc.2`, SemVer spelling). `CITATION.cff` stays at `1.46.0` — the last GA — per the prerelease convention the packaging guard enforces. CHANGELOG stays under the final `[1.47.0]`: the rc soaks for it.

Packaging smoke 16/16.

### DIFF
--- a/packaging/helm/distil-gateway/Chart.yaml
+++ b/packaging/helm/distil-gateway/Chart.yaml
@@ -7,7 +7,7 @@ type: application
 # Chart version tracks the chart; appVersion tracks the distil release it defaults to.
 # Bump `version` on any template change, `appVersion` on a distil upgrade.
 version: 0.1.1
-appVersion: "1.47.0rc1"
+appVersion: "1.47.0rc2"
 home: https://dshakes.github.io/distil/
 sources:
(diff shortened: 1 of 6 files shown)


dshakes:
Superseded: cutting the final 1.47.0 directly (maintainer decision).

PR #136, opened 2026-08-15
chore(homebrew): point the formula at v1.47.0

dshakes:
`scripts/release.sh` patches `packaging/homebrew/distil.rb` in place after pushing the tag — it fetches the release tarball and rewrites the `url` and `sha256`. This commits that edit so the in-repo formula matches what shipped.

The formula was still on `v1.40.1`, which is expected: the tap is canonical and this file lags until a release refreshes it.

Verified the sha against the actual tarball rather than trusting the script — see below. Packaging-metadata only; nothing a running proxy/wrap/gateway executes.

### DIFF
--- a/packaging/homebrew/distil.rb
+++ b/packaging/homebrew/distil.rb
@@ -8,16 +8,16 @@
 #   brew tap dshakes/tap
 #   brew install dshakes/tap/distil
 #
-# sha256 is for the v1.40.1 source tarball. To recompute for a new version:
+# sha256 is for the v1.47.0 source tarball. To recompute for a new version:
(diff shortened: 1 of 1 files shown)


dshakes:
Verified the sha256 independently against the published tarball rather than trusting the driver's own output:

```
formula sha: 91bada5712b98210baa15cf54e6fe06e21a4a4ee12cb1a355aab9eab7ff05cba
tarball sha: 91bada5712b98210baa15cf54e6fe06e21a4a4ee12cb1a355aab9eab7ff05cba
MATCH ✓
```

(`curl -sL .../v1.47.0.tar.gz | shasum -a 256`)

PR #147, opened 2026-08-20
test(audit): guard the unreadable-trail assert against root

dshakes:
## What

Adds the two skip guards (`win32` + `geteuid()==0`) to `tests/test_audit.py::test_unreadable_trail_reads_as_empty`.

## Why

`chmod 0o000` is a no-op for root: the trail stays readable, so `read_events()` returns the recorded row instead of `[]` and the test fails. CI runners are non-root, so this never fires in CI — but a **root** gate run (e.g. a containerized green-keeper) hits a false failure on an otherwise-green tree.

The identical `chmod 0o000` test in `tests/test_doctor.py` (`test_check_mode_unreadable_file`) already carries both guards verbatim; this mirrors them. The assertion is unchanged where it can hold — nothing is weakened, no product code changes.

## Verification (local, run as root)

- `uv run --with pytest --with pillow python -m pytest -q` → **3448 passed, 10 skipped** (the previously-failing test is now the 10th skip)
- `uvx ruff check distil tests` → clean; `uvx ruff format --check .` → clean
- `uvx mypy --python-version 3.12 --ignore-missing-imports distil` → clean
- `uv run distil bench` / `verify` / `validate` → all **PASS**

## Scope

Test-only, 5 insertions. No thresholds, gates, versions, tags, or workflows touched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N3v51W3tz3KWqcLP3KHykR)_

### DIFF
--- a/tests/test_audit.py
+++ b/tests/test_audit.py
@@ -199,6 +199,11 @@ def test_reader_handles_blank_lines_and_a_missing_file(tmp_path) -> None:
     sys.platform == "win32",
     reason="chmod 0000 does not make a file unreadable on Windows, so the failure path cannot be provoked",
 )
+@pytest.mark.skipif(sys.platform == "win32", reason="chmod 0o000 is a no-op on Windows")
+@pytest.mark.skipif(
+    hasattr(os, "geteuid") and os.geteuid() == 0,
+    reason="chmod 0o000 is a no-op for root (bypasses permission bits)",
+)
 def test_unreadable_trail_reads_as_empty(tmp_path) -> None:
     """A trail that cannot be opened must not raise into the CLI."""
     import os
--- a/tests/test_audit.py
+++ b/tests/test_audit.py
@@ -195,26 +195,29 @@ def test_reader_handles_blank_lines_and_a_missing_file(tmp_path) -> None:
     assert [e["key_id"] for e in audit.read_events()] == ["gk_1"]
 
 
-@pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


dshakes:
Good catch — this test already carried a win32 skip before this branch touched it, so my first commit duplicated it instead of just adding the root guard. Fixed in ec8c76c: removed the redundant second `skipif`, kept the original win32 decorator plus the new `geteuid()==0` guard. Re-verified locally: `tests/test_audit.py` → 11 passed, 1 skipped; ruff check/format clean.

---

dshakes:
Adopted the Codex suggestion in 71f80bc — it's the better fix, and it makes the PR strictly additive rather than a coverage trade.

Instead of skipping under root, the test now raises `PermissionError` from `Path.open` for the audit path. That provokes `read_events()`'s `except OSError` deterministically on every platform and every uid, so the assertion runs *everywhere* — including root containers and Windows, where the original `chmod 0o000` version was a silent no-op that could never fail.

Net effect on the suite: **3449 passed / 9 skipped** (baseline on main was 3448 / 10) — one more test actually executing, and both skip decorators are gone.

Verified locally as root: full suite green, `ruff check` + `format --check` clean, `mypy --python-version 3.12` clean, and `distil bench` / `verify` / `validate` all PASS.

Note for the record: the Gemini audit posted at 18:46 was reviewing the earlier commit (the skip-based version) and is now superseded.

---

PR #158, opened 2026-09-04
fix(onboard): Ctrl+C inside the launched agent no longer tears onboard down

dshakes:
`distil onboard` hands the terminal to `distil wrap -- claude` with a plain
subprocess.run. The terminal delivers SIGINT to the whole foreground group, and
Claude Code uses the first press to cancel a turn, not to exit — so the same key
raised KeyboardInterrupt in onboard's parent and dumped a traceback over the
still-running agent.

wrap's own parent already handles this with a no-op Python-level SIGINT handler
(proxy.py); onboard now installs the same one before the handover. A Python-level
handler resets across exec, so the child still receives its Ctrl+C.

The existing --yes step-1 test now asserts the handler is installed; it fails on
the old code.

## Verification
- `tests/test_onboard.py::test_cmd_onboard_yes_runs_first_step` extended; fails without the fix, passes with it
- `tests/test_onboard.py tests/test_cli_onboard.py`: 95 passed
- ruff@0.15.10 check + format clean

### DIFF
--- a/distil/cli.py
+++ b/distil/cli.py
@@ -2063,6 +2063,15 @@ def cmd_onboard(args: argparse.Namespace) -> int:
         first_cmd = steps[0][1]
         if ask(f"Start now — run step 1?  ({first_cmd})"):
             print()
+            # Ctrl+C belongs to the agent (it cancels a turn, not the session).
+            # Same rule as wrap's own parent — see proxy.py's SIGINT note. A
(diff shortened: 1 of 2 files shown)


dshakes:
Taken: the no-op handler is now scoped to the `subprocess.run` call with try/finally, and the test asserts both the no-op during the handover and the caller's handler restored afterwards.

PR #172, opened 2026-09-07
feat(cache): the client rewrites its history every turn, and it cost us the prefix

dshakes:
ADR 0008 promised byte-stable forwarding *when the client re-sends a message byte-identical*, and clause (d) said a client that rewrites its own history gets no promise. That clause was written as an edge case. It is the common case.

Agentic clients rewrite their history on **every turn** without changing a token the model reads: the `cache_control` breakpoint advances to the newest block, an SDK shim stamps positional `index` fields, a string becomes a single text block and back again. distil forwards what it receives, so all of it reached the provider as changed bytes and missed a prefix the provider was already holding — at the write rate, silently, with the bill as the only symptom.

## What it cost, measured

`benchmarks/prefix_replay_stability.py` — offline, deterministic, no API calls. Share of re-sent messages forwarded byte-identical to the previous turn, 8 turns:

| provider shape | rewrite | before | after |
|---|---|---|---|
| anthropic/messages | `index` stamped | 0.0% | 100.0% |
| anthropic/messages | string/block sugar | 0.0% | 100.0% |
| openai/chat-completions | `index` stamped | 14.6% | 100.0% |
| openai/chat-completions | string/block sugar | 0.0% | 100.0% |
| openai/responses | `index` stamped | 0.0% | 100.0% |
| openai/responses | string/block sugar | 0.0% | 100.0% |
| gemini/generateContent | `index` stamped | 0.0% | 100.0% |

Zero, not merely degraded. A client doing nothing but renumbering an `index` field lost the entire prefix, on every provider, on every turn. The un-rewritten control reads 100.0% before and after, so replay repairs rewrites and does nothing when there is nothing to repair. Added latency ~0.13 ms/request.

Two things this does **not** claim. It measures what distil forwards, not what the provider billed — live confirmation is the cache-read share in `distil dissect`, and that needs a soak after release. And the marker-advance case was already free: anchoring the recency carve-out to the client's breakpoint (ADR 0008) neutralised it, and the same table reads 100% for it before and after. That is a result, not a gap.

## The mechanism

Per conversation lineage (leading system-run + model + tools + canonicalised head, per session), distil remembers the previous turn's `(original, forwarded)` pair. For the longest canonically-equal leading prefix it forwards **the bytes it forwarded last turn**, re-placing the client's *current* `cache_control` markers at the client's *current* block positions, and compresses only the divergent suffix. Stop at the first divergence and stay stopped — a cached prefix is a byte prefix.

The comparison ignores `cache_control`, `index`, the interchangeable spellings of a single text block (bare string / `text` / `input_text` / `output_text`), and JSON key order. It ignores nothing else. Tool inputs are opaque: compared as they arrive, never structurally rewritten, because a tool input can itself contain a key called `content`. `citations`/`annotations` are deliberately excluded — the cost of leaving them out is a missed hit, the cost of putting them in wrongly is a wrong prefix.

State is in-memory, LRU-bounded (16 lineages, 8 MB of history each at most), and **not persisted**. It holds message bytes, and the TTL'd restore store is the only place content is allowed to rest on disk. A hot-swap starts a lineage cold and rebuilds from the next turn, at a cost of one cache write. Fail-open: any exception forwards exactly what the compressor produced.

On by default per the house rule; `--no-prefix-replay` on `wrap`, `proxy` (sync and async) and `gateway`, plus `gateway.prefixReplay` in the Helm chart. An operator who can turn a thing on in one place and not off in another does not really have a flag.

## The guard, which is the part that isn't obvious

distil's compressor is **not** a pure function of one message. The exact-quote guarantee (`compress.provenance`) keeps a tool result verbatim because of an `Edit` that arrives *later* in the list, so a block legitimately digested at turn N can need to be verbatim at turn N+1. Overlaying turn N's stub there would break the agent's next edit to buy a cache hit.

So a message is replayed only when **this turn's forwarded form is itself canonically equal to last turn's**. Replay restores bytes, never decisions. That makes "prefix replay never changes semantic content" the loop condition rather than a claim about the implementation — and `distil validate` asserts it as a sixth invariant anyway, because a loop condition is exactly what a later optimisation deletes.

## Proof

`tests/test_cache_contract.py` gains three rewrite shapes × four request shapes, each with a **control arm**: the test fails unless the rewrite demonstrably shortened the byte-stable prefix *without* replay and demonstrably did not *with* it. A stability assertion whose control never destabilises proves nothing.

Plus: a semantic edit breaks the prefix at exactly its index; a changed `tool_result` is never overlaid with older bytes; a re-signed `thinking` block breaks the prefix while an unchanged one is replayed byte-for-byte; the lineage never crosses a model/tools/system change; state is bounded and an oversized history is released rather than held; the fail-open paths forward unchanged. The canonicalizer's strip set is pinned from both sides — a dropped `else` in its recursion made every message canonicalise to `{}` during development, and the parametrized tests caught it only by luck.

End to end through the real proxy against a stub upstream: the bytes the upstream receives are unchanged across a churned turn, paired with the same session run with the feature off to show the churn really does move bytes.

## Measurable

Per-request record gains content-free `replay_hits` / `replay_misses` / `replay_restored`; `distil dissect` reports them per session beside the cache-read share. `restored` is never folded into `hits` — hits with zero restored is the healthy steady state, and adding them together would make a dead mechanism look identical to a working one, the same failure ADR 0008 called out for `None` vs `0.0%`. `None`, not zeros, for records that predate this.

## Prior art

Named in ADR 0011: Headroom's `PrefixCacheTracker.overlay_cached_prefix` is the same mechanism, with the same comparison-key/forwarded-bytes separation, opaque tool inputs, breakpoint re-placement and lineage scoping. It is the right mechanism and there is no point pretending otherwise. What differs is the guard above and the fact that this one is falsifiable.

## All three servers, and whose credential it is

The threaded proxy, the async proxy (`--async`) and the multi-tenant gateway apply replay at the same point through one shared `prefixreplay.apply`, so the fail-open exists once instead of three times. A default-on feature that only one server has is a feature its users do not have, and managed installs run `distil proxy`.

Each scopes the lineage by whose credential it is, because the lineage key is otherwise content-derived: the gateway per tenant, the plain proxies by a hash of the client's own API key. Both are asserted by a test that posts the identical conversation under two identities and fails if the second is served the first's bytes.

## Since the first review round

Four findings from the cross-audits, each reproduced before it was fixed and each left behind a test that fails on the unfixed code:

1. **A tool result's own `index` is data.** The comparison key stripped `index` at every depth, so Gemini's `functionResponse.response` — arbitrary tool output — made `{"index":1,"value":"A"}` and `{"index":2,"value":"A"}` the same key and the second request went out with the *first* result's bytes. Tool payloads are now opaque in both directions, arguments out and result back.
2. **Replay wrote back into the state it replayed from.** `_reput` returns the stored item untouched when the message-level marker has not moved, and the block pass then assigned `content` into it — writing this turn's marker placement into state a concurrent request on the same conversation is about to read. Copy on write, key order preserved.
3. **A breakpoint with nowhere to sit.** Client sends a bare string one turn and one text block *with* a `cache_control` breakpoint the next; the comparison calls those equal, and replaying the string dropped the marker. On Anthropic the breakpoint is the cache entry, so that buys the hit by destroying the thing being hit. Replay stops at that index instead.
4. **Credential scope on the plain proxies.** The gateway scoped by tenant; the sync and async proxies scoped by nothing. What could cross was byte *spelling* of a canonically-equal history rather than content, but the boundary already existed one server over.

Also: a `--no-prefix-replay` session now records **no** counters rather than three zeros, so `dissect`'s *not recorded* means what its docstring says. And the cache contract renumbers this amendment to clause **(f)** — ADR 0010 (the re-read delta, #173) took (e) on main.

The re-read delta and replay are now driven together through the real proxy, since they meet on the same contract: a stub emitted at turn N is forwarded byte-identical at turn N+1 through the client renumbering its `index` stamps, and when a later `Edit` straddles the elision cut the quote guard withdraws the stub — so the block goes out verbatim and replay does **not** overlay last turn's bytes.

## Gates (run locally, exact CI commands)

- `ruff check distil tests` — pass
- `ruff format --check .` — 373 files already formatted
- `mypy --python-version 3.12 --ignore-missing-imports distil` — no issues, 141 source files
- `pytest -q --cov=distil --cov-fail-under=95` (py3.12, CI extras) — **3931 passed, 13 skipped, 95.13%**
- `distil validate` — PASS 175/175 over 25 cases, including `prefix-replay-semantics`
- `distil bench` — PASS, every trajectory certified non-inferior
- `distil verify` — PASS
- `distil retention --max-lost 0` — 0 lost
- `distil fidelity --max-silent 15` — 9
- `pytest tests/test_packaging_smoke.py` — 16 passed

`docs/changelog.html` and `docs/search-index.json` are regenerated by their scripts, as their gates require.

Unverified: no live traffic touched this. The provider-side effect is a claim about forwarded bytes only until `distil dissect` reports a cache-read share from a real soak.

Rebased onto `0153ff7` (the re-read delta, #173). One `## [Unreleased]` above `## [1.52.0]`, carrying #173's entry and this one — the duplicated copy of the released 1.52.0 section is gone. `benchmarks/prefix_replay_stability.py` re-run after every fix above: the table is unchanged, every applicable pairing still 100.0%.

Not merged.


dshakes:
The BLOCKING verdict says the gateway's per-tenant scoping has zero test coverage. It has coverage, and the coverage is load-bearing — verified by mutation rather than by reading.

The test is `test_the_gateway_holds_the_prefix_and_never_shares_it_between_tenants` in `tests/test_cache_contract.py` (it lives with the other cache-contract tests, not in `test_gateway.py`, which is probably why a grep missed it). It stands up the real gateway handler against a stub upstream and posts the same conversation as two tenants.

Removing `scope=tenant` from `distil/gateway.py` makes it fail:

```
AssertionError: cross-tenant leak: globex sent index=3 and the gateway forwarded
index=1, which is acme's — the tenants are sharing replay state
1 failed, 65 deselected
```

The reviewer's confusion was fair on one point: the original assertion was `index == 3`, which only reads as an isolation check once you have reconstructed why a shared lineage would make it 1 (the canonical comparison ignores `index`, so the two tenants' requests are canonically equal and the second would be overlaid with the first's stored bytes). Pushed `4a57656`, which keeps the same test and the same mutation coverage but states the leak in the failure message instead of leaving it to be inferred.

No production code changed.

PR #173, opened 2026-09-07
feat(compress): a re-read need not re-send the lines the first read delivered

dshakes:
The coding hot path is read → edit → **re-read**, and it is where the exact-quote
guarantee is most expensive. Content the agent may have to reproduce in an
`Edit(old_string=…)` is kept byte-exact forever, and under a client that caches its whole
history — which is what Claude Code does — a superseded read cannot be demoted either, so
the bytes accumulate. This recovers some of that without touching the guarantee.

## What the traffic looks like

Measured over 2,489 local Claude Code sessions (52,937 tool results, 19.4M tool-result
tokens, content-free):

| | |
|---|---|
| `Read`/`cat` results that re-read a path already read this session | **51.4%** |
| of those, byte-identical to the earlier read | 12.8% |
| of a re-read's tokens, lines already delivered verbatim earlier in the conversation | **50.6%** |

Half the mass of a re-read is a second copy of lines the model is already looking at.
Neither existing mechanism could see it: exact dedup needs identical bytes, and
`cachedelta`'s near-duplicate gate compares whole blocks at a 0.5 `difflib` ratio, which
misses **608 of 726** changed re-reads — two disjoint windows onto one file are not 50%
similar *as blocks* even when every line they share is byte-identical. It fires on 5.2% of
re-reads and removes 0.23% of tool-result mass.

The unit was wrong. Block similarity asks "is this nearly the same document"; the useful
question is "which of these lines have I already sent".

## What this does

The longest run a read shares with an earlier read of the same path becomes a reversible
reference; every other line stays verbatim.

```
«distil-reread handle=a1b2c3d4» lines 1-20 of this result are byte-identical to
lines 41-60 of the earlier read of /app/handlers.py, which is still in this
conversation verbatim. Call distil_expand with this handle to recover them here.
```

**Why the guarantee still holds.** It is a statement about the *conversation*, not about
any one block: an `Edit` applies if its quote occurs byte-exact anywhere in the forwarded
payload, and an agent quoting lines it saw can be served by either copy. Five rules:

- **The freshest tool output is never elided.** The recency carve-out applies here like
  everywhere else — a re-read the agent has just issued is precisely the output it reasons
  over to choose its next action. The *plan* is computed from the message prefix
  regardless, so which blocks may be bases never depends on the sliding window; only the
  application is gated.
- **Only a name-keyed read may be a base** (`Read`, `view`, `read_file`, …). Those are
  exempt unconditionally, so the referenced lines cannot leave the conversation later. A
  shell `cat` may be a *target* but never a base — its exemption is conditional on not
  being superseded, and a whole-file shell re-read supersedes exactly the block it would
  want to point at. Pinning it would forfeit the base's digest to save the same bytes on
  the copy, a wash at best.
- **References never chain.** An elided block is not itself a base.
- **Cuts interior to the block are pulled back 20 lines**, so a quote straddling the
  boundary still occurs contiguously in the base. A run reaching the block's own first or
  last line takes no margin there.
- **Runs under 8 lines are left alone**, and **lines carry their terminators into the
  comparison** so a CRLF base and an LF re-read never match.

Reversible through the same `RestoreStore` as every other stub.

## Cache safety — and a deliberate deviation from ADR 0008 (b)

There is **no volatile-suffix gate**, and that is the point rather than an oversight.

The plan is a pure function of the message *prefix*, so which lines a block may drop never
changes as the conversation grows — the same construction `cachedelta` relies on. A
boundary gate would have been strictly worse: ADR 0008 records that under the Claude Code
shape the entire history is cached and there is no uncached tail, so the transform would
never run on the only traffic it targets; and it would *introduce* the failure clause (a)
exists to catch, flipping a block from stub to verbatim exactly as the boundary advances
past it.

The recency carve-out sits on top, and because that carve-out is itself anchored to the
breakpoint, the one rendering change it causes lands strictly after the boundary — exactly
as clause (d) already describes for every other transform.

The cache contract gains clause **(e)** stating prefix determinism as the second way to
satisfy (a). `tests/test_reread_delta.py` asserts byte-stability under the moving-marker
(fully-cached) shape, plus a test that fails if the delta stops firing. ADR 0008 is
annotated as amended by ADR 0010. **Flagging this for review** — it is the one design call
here that reads against the letter of an accepted ADR, and the argument is in
`docs/adr/0010-the-re-read-delta.md`.

## Measured

On the codebench corpus (read → edit → re-read, 20 sessions / 320 turns), under the client
shape that actually bills — newest turn pinned, whole history cached:

| `distil (PAYG digest)` | before (`4cf076b`) | after |
|---|---|---|
| token savings | 31.4% | **41.7%** |
| cache-aware dollar savings | 35.8% | **49.1%** |

`distil bench` is **unchanged byte for byte** — its corpus has no re-read of one path
through a name-keyed read tool, so the delta never fires there.

**codebench is nothing but reads, so it overstates.** On real traffic name-keyed reads are
10.7% of tool-result mass, about half are re-reads, and about half of a re-read's tokens
were already delivered — a live ceiling near **2.8% of tool-result mass**, in line with the
~2.3% the investigation projected. That is the number to quote.

### One benchmark caveat this surfaced, stated rather than hidden

`benchmarks/codebench.py` builds its sessions with **no** `cache_control` marker and then
prices them **with** a cache — the longest stable prefix is billed at the cache-read rate.
No Anthropic client looks like that: Anthropic caches only what the client marks, so an
unmarked request has no cached prefix at all, and every recency-anchored carve-out distil
has is charged there for busting a cache that was never created.

| client shape | before | after (tok) | after ($) |
|---|---|---|---|
| newest turn pinned (Claude Code) | 31.4% / 35.8% | **41.7%** | **49.1%** |
| no marker (codebench as shipped) | 0.0% / 0.0% | 7.1% | **−15.4%** |

The −15.4% is left visible. The artefact predates this PR — `distil-verbatim` shows 18.5%
tokens for 3.7% dollars on the same corpus — and was invisible only while the digest row
was 0.0% and nothing moved. New runner `benchmarks/codebench_marked.py` replays both shapes
so it is reproducible rather than asserted. **The corpus itself is left alone**: changing it
would move the competitor rows on `docs/compare.html` too, which is out of scope here.

Planner cost is 0.053 ms/turn over the same 320 turns. The digest row's `ms/turn` rise
(≈1 → ≈6) is the on-disk restore store, which every stub pays and which that row had never
paid before because it emitted no stubs on this corpus.

Raw before/after stdout: `benchmarks/results/2026-09-06/`. Competitors were not installed
in that environment, so those rows are absent from both `codebench` files — it is a
distil-vs-distil comparison and `docs/compare.html` is untouched.

## Codex quote hazard (the 1.51 follow-up)

1.52.0 shipped the exemption on all three adapters but the *measurement* on the Messages
path only, so Codex traffic reached `distil dissect` reported as carrying no edits at all.

`apply_patch` is the gap: on GPT-5 models it is a **freeform** custom tool whose argument
is the patch body, not JSON, so there is no `old_string` field. What must match the file is
each hunk's **pre-image** — the context and removed lines in order, running *through* the
`+` lines, since an added line is not part of what the patcher looks for.
`provenance.patch_quotes` extracts that per the tool's own Lark grammar;
`response_edit_quotes` normalises both `custom_tool_call` and `function_call`;
`observed_view` now drops model-output *item types* as well as assistant messages, without
which the envelope quotes the file to itself and the check could only pass. A one-line hunk
counts only when the line is long enough to discriminate — dropping them all blinds the
gauge, counting a one-line `}` books a false survivor. The Responses path gets the same
widen-on-miss reaction. Nothing downstream changed — both adapters already share one
thread-local counter.

## Review round 1

All three findings from the first CI round are addressed in `e2e661b`; see
[this comment](https://github.com/dshakes/distil/pull/173#issuecomment-5564292683) for the
detail. The `local_shell_call` finding turned out not to be reachable — only
`function_call_output` is compressed on the Responses path — so it is pinned with a test
rather than fixed with speculative code.

## Gates

| gate | result |
|---|---|
| `ruff@0.15.10 check distil tests` | All checks passed |
| `ruff@0.15.10 format --check .` | 371 files already formatted |
| `mypy@1.17.1 --python-version 3.12 --ignore-missing-imports distil` | no issues, 140 files |
| pytest py3.12 + CI extras, `--cov-fail-under=95` | **3879 passed, 5 skipped, 95.11%** |
| `distil validate` | **PASS — 150/150 across 25 cases** (4 new re-read shapes) |
| `distil bench` | PASS, unchanged aggregate |
| `distil verify` | PASS |

New module is at 100% line coverage. Rebased onto `5351b61`; the CHANGELOG entry sits in a
fresh `## [Unreleased]` above `[1.52.0]`.


dshakes:
Pushed fixes for all three review findings, plus a benchmark artefact they turned up.

**`review` (blocking) — recency carve-out bypassed.** Correct, and now fixed. The elision is gated on `is_recent`, so the freshest tool output goes out byte-exact; a re-read the agent has just issued is exactly the output it reasons over to choose its next action. The *plan* is still computed from the message prefix, so which blocks may serve as bases never depends on the sliding window — only the application is gated, and where a client marks a cacheable prefix that window is empty or lies wholly after the boundary. Pinned by `test_the_freshest_tool_output_is_never_elided`.

**`audit` (blocking) — CRLF vs LF.** Correct, and sharp. The planner compared `splitlines()` while the applier removed bytes with `splitlines(keepends=True)`, so a CRLF base and an LF re-read matched and the stub's claim of byte-identity would have been false. Both sides now use `keepends=True`, which makes a mixed-terminator pair simply not match. Pinned by `test_a_crlf_reread_does_not_match_an_lf_base` and, end to end, `test_the_elided_bytes_are_exactly_what_the_base_holds`.

**`audit` (blocking) — `local_shell_call` not scanned.** Not reachable, and I checked rather than assumed: `_compress_response_item` compresses only `function_call_output`. A `local_shell_call_output` is passed through untouched, so its bytes always survive and there is nothing to exempt. Rather than add code for a path that does not exist, I pinned the reason it is safe — `test_only_function_call_output_is_compressed_on_the_responses_path` fails the moment another output type gains a compressor without its exemption.

**`audit` (non-blocking) — single-line pre-images.** Fair. A one-line hunk now counts when the line is long enough to discriminate. Dropping them all blinded the gauge exactly where it is meant to see Codex; counting a one-line `}` would book a false survivor on a request where nothing survived. Under-counting is silence, over-counting is a false all-clear on a safety gauge.

**Gemini (non-blocking) — `base_ok` clarity.** Taken: `FILE_READ_TOOLS` is exported and `base_ok` keys on it directly instead of relying on `read_target` to filter `grep`/`glob` out downstream.

---

### The numbers changed, and chasing that turned up a corpus artefact

Honouring recency moved the measurement, so I re-ran it — and found that `benchmarks/codebench.py` builds its sessions with **no** `cache_control` marker and then prices them **with** a cache. No Anthropic client looks like that: Anthropic caches only what the client marks, so an unmarked request has no cached prefix at all, and every recency-anchored carve-out distil has is charged there for busting a cache that was never created.

`distil (PAYG digest)`, by client shape:

| client shape | before (`4cf076b`) | after (tok) | after ($) |
|---|---|---|---|
| newest turn pinned (Claude Code) | 31.4% / 35.8% | **41.7%** | **49.1%** |
| no marker (codebench as shipped) | 0.0% / 0.0% | 7.1% | **−15.4%** |

The −15.4% is left visible rather than hidden. The artefact predates this PR — `distil-verbatim` shows 18.5% tokens for 3.7% dollars on the same corpus — and was invisible only while the digest row was 0.0% and nothing moved. New runner `benchmarks/codebench_marked.py` replays both shapes so it is reproducible rather than asserted. **The corpus itself is left alone**: changing it would move the competitor rows on `docs/compare.html` too, which is out of scope here.

The site, ADR, CHANGELOG and `claims.json` now quote the marked pair (31.4% → 41.7% tokens, 35.8% → 49.1% dollars), with the live ceiling still stated as ~2.8% of tool-result mass. Rebased onto 1.52.0 and the entry lifted into a fresh `## [Unreleased]`.

Gates re-run locally: ruff check + format clean, mypy clean, **3879 passed / 5 skipped at 95.11% coverage** on py3.12 with the CI extras, `distil validate` 150/150 across 25 cases, `distil bench` PASS and unchanged, `distil verify` PASS.

PR #177, opened 2026-09-07
fix: Windows file-lock fallback + OpenAI/Gemini vision-dedup parity

dshakes:
## Summary

Two platform/provider parity gaps closed on one branch, both "works on the primary path, silently worse on the others" defects a same-platform, single-provider test run cannot surface:

- **Windows file locking** — `gateway_keys.py`, `audit.py`, `ledger.py`, `census.py`, `shadow.py`, `retention.py`, `surfaces.py`, and `mcp_server.py` each guarded `import fcntl` with a bare `ImportError` and fell back to **no lock at all** on Windows. New `distil/_filelock.py` (`fcntl.flock` on POSIX, `msvcrt.locking` against a `.lock` sidecar on Windows, fail-open on any platform-call error) is now the one place any of them ask for mutual exclusion. Previously win32-skipped tests now exercise the Windows branch directly (a fake `msvcrt` module), so they run on every CI runner instead of only a Windows one.
- **Vision duplicate elision (ADR 0003) was Anthropic-only** — the certificate-gated dedup that replaces a repeated image with a reversible reference never covered OpenAI's `image_url`/`input_image` parts or Gemini's `inlineData`/`fileData` parts, so those adapters paid full price on every repeat of an already-sent image. The identity/elision logic now lives once in `compress/vision.py` (`elide_or_keep`), shared by all three adapters behind the same certificate gate and the same `image_kept`/`image_elided` census reasons. `proxy._count_messages` gained the matching `image_url` token count so the new OpenAI elisions score their savings correctly instead of clamping to zero.

## Test plan

- [x] `ruff@0.15.10 check` — clean (whole repo)
- [x] `ruff@0.15.10 format --check` — clean (whole repo, 377 files)
- [x] `mypy@1.17.1 --python-version 3.12 --ignore-missing-imports distil` — clean (142 files)
- [x] `uv run --with pytest --with pytest-asyncio --with aiohttp --with pytest-cov --with opentelemetry-sdk --with cryptography --with pillow python -m pytest -q --cov=distil --cov-fail-under=95` — **3967 passed, 13 skipped, 95.10% coverage**
- [x] `distil bench` — GATE: PASS (every trajectory certified non-inferior)
- [x] `distil verify` — FIDELITY: PASS (reversible + append-only across the corpus)
- [x] `distil validate` — VALIDATE: PASS (175/175 checks, 25 adversarial cases)
- [x] New tests: `tests/test_filelock.py`, `tests/test_vision_openai.py`, `tests/test_vision_gemini.py` (mirror `test_vision.py`'s Anthropic coverage per new shape, plus a cache-contract test that elision output is byte-stable across separate calls)


dshakes:
Superseded by the new PR on fix/platform-parity-2: identical commits rebased onto main after #176.

PR #187, opened 2026-09-16
feat(reach): three more agents wrapped, sixteen declined with their reasons on record

dshakes:
## What
`distil wrap` reaches **18 agents** (was 15). The competitor set claims 17–20; the difference is now documented target by target rather than asserted.

**Rule applied throughout:** a new preset lands only when its routing contract is verified from the tool's primary docs or source (URL + date in the target's docstring). A wrong config shape breaks a user's agent, so anything unverifiable was declined, not guessed.

**Added**
- **Mistral Vibe** (env-var): `VIBE_PROVIDERS` is a JSON provider *array*, verified from mistral-vibe's ADR 0005 + `config/layers/environment.py`. Exporting a bare URL would route nothing while reporting success, so `wrap_run` gained an `env_value_template`; one merging element redirects `mistral` and touches no file.
- **Cline CLI** (config file): `~/.cline/data/settings/providers.json`, `providers.<id>.settings.baseUrl` (+ `lastUsedProvider`, which selects the active provider), honours `CLINE_DATA_DIR`; schema from the zod `StoredProviderSettings` + a committed fixture.
- **Kilo Code CLI** (config file): `~/.config/kilo/kilo.json`, `provider.<id>.options.baseURL`; a JSONC file is refused untouched with the one line to add by hand, never rewritten without its comments.

Both config-file targets use the existing claim → backup → patch → restore mechanism with the standard round-trip tests (absent/present/idempotent/no-credential/unparseable).

**Declined, with the mechanism (recorded in `distil/targets.py`)**
No base-URL knob: Cursor CLI, Amp, auggie, Antigravity, Tabnine · knob cannot be local: Warp, Cody · Secret Storage, not a file: Roo Code · no single key + app rewrites the file while running: Zed · shared daemon serving other users: OpenClaw · desktop app, no process to scope: ZCode · no plain file key: VS Code Copilot, Cortex Code · docs unreachable: Trae, Junie.

**Also:** `distil wrap --list` / `--json`; README + `docs/integrations.html` + `docs/IDE-AGENTS.md` tables generated by `scripts/build_agent_tables.py` with a drift test; stale Warp/Cline notes fixed; Azure `/openai/v1/` `api-version` is GA and optional; `COPILOT_MODEL` documented. The architecture diagram and its `claims.json` entry now say 18.

## Verified locally (exact CI commands)
pytest full **4089 passed, 13 skipped**, coverage **95.14%** (floor 95) · ruff, mypy clean · agent tables in sync · check_nav in sync.


dshakes:
Update after review (merged current main in first — #182/#183/#185/#186):

**Same defect class, fixed where the client's source was read** — clients that use a configured base URL literally (append only the leaf path) were routed to a path the proxy doesn't compress:
- **kilo** (`@ai-sdk/anthropic` + `@ai-sdk/openai`) → `$BASE/v1`
- **aider** (LiteLLM `OPENAI_API_BASE`), **opencode** (`@ai-sdk/openai`), **qwen** (`openai` npm), **openhands** (`LLM_BASE_URL` → LiteLLM) → `$BASE/v1`
- **grok** (`xai-org/grok-build` `resolve_inference_base_url`) and **kimi** (`kimi-code` `kimi.ts`: `options.baseUrl ?? env.KIMI_BASE_URL ?? 'https://api.moonshot.ai/v1'` → `new OpenAI`) — were a silent no-op (forwarded uncompressed); upstream default loses `/v1`, template gains it, in the same commit.
- codex/opencode shape relabelled to OpenAI Responses.
- **Unverified, unchanged:** goose, copilot. **Codex** may not route through `OPENAI_BASE_URL` at all (native Rust client, base URL from TOML config) — documented, needs its own follow-up.

**Contract tests** (`tests/test_reach_contract.py`): every templated preset renders its env, sends each SDK's real request shape through the real proxy to a fake upstream, and asserts the path is compressible, forwarded correctly, and that **compression actually ran** (`x-distil-compressed == "1"`, tokens saved > 0). Reverting the Kilo fix fails `kilo-anthropic`/`kilo-openai`. A tautological test was removed; `/v1/v1` behaviour pinned (allowlist untouched).

Two independent reviews: no blockers; no persistence of preset upstreams, so no mismatch for existing users. I spot-checked the kimi source citation against the live repo.

Full CI gate on be97097 (py3.12): ruff ✅ · mypy ✅ · pytest --cov-fail-under=95: 4574 passed ✅ · packaging ✅.

